Issue #11, opened 2021-01-07
Add annex's uuid to URL model


yarikoptic:
Relates if not partially replaces (for the use case of determining if a URL might be providing annexed data) #10 . 
Model has `dsid` which is DataLad's `Dataset.id`.  Every annex repository/remote has a UUID.  We should record those two whenever a URL (remote) has UUID.  That would allow to judge either any given URL should be considered as a source of data (motivation behind #10 as well) 

PS why not to rename `dsid` into `ds_id` and then here use `annex_uuid`.  Although longer -- it IMHO would make it more "obvious"

PR #14, opened 2021-01-18
Use 'ds_id' rather than 'dsid'

kyleam:
The separation should make it more obvious that this is referring to
the dataset ID.

Suggested by @yarikoptic in gh-11.

---

Or should we just go with a full `dataset_id`?


kyleam:
> Or should we just go with a full `dataset_id`?

Sticking to the shorter variant for now.

PR #16, opened 2021-01-18
tasks: Collect annex.uuid of source repository

### LINKED ISSUE #11: Add annex's uuid to URL model

kyleam:
Closes #11.

### DIFF
--- a/datalad_registry/factory.py
+++ b/datalad_registry/factory.py
@@ -28,8 +28,8 @@ def setup_celery(app, celery):
     }
     cache_dir = app.config.get("DATALAD_REGISTRY_DATASET_CACHE")
     if cache_dir:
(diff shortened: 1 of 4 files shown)


yarikoptic (approved):
FWIW, looks legit

kyleam:
Thanks for taking a look.

PR #31, opened 2021-04-29
Add typing hints to non-test code

kyleam:


kyleam:
> should then https://github.com/datalad/datalad-registry/blob/master/.github/workflows/test.yml get a mypy run?

As I mentioned in the second commit, I think local for now is fine, but I can add it in this PR if you prefer.

yarikoptic:
If it passes locally, I see no reason to not add it to CI

PR #43, opened 2021-08-24
Add git_objects_kb field

### LINKED ISSUE #40: add size of .git/objects to the schema/display
even though it is subject to fluctuate depending on local repacking/gc, I think it would be useful to gather/display that information since would be reflective of space etc needed for an initial clone of that repository

jwodder:
Closes #40.


yarikoptic:
Great, thank you!

PR #44, opened 2021-08-25
Setting up environments

### LINKED ISSUE #42: "Environments"
In response to https://github.com/datalad/datalad-registry/pull/39#issuecomment-904882103 by @jwodder 

> What exactly are your requirements regarding environments for development, testing, and/or production and how they're set up?

For all cases, let's 
- make specification of `DATALAD_REGISTRY_PASSWORD` env var mandatory (crash if not specified) and use it for postgres and any other component which requires authentication (I see `RABBITMQ_DEFAULT_PASS`...).  For CI - could be set to an arbitrary value.

- testing: as it is now (as e.g. in #39) with a fixture taking care about docker compose up/down, just make it possible (if tests allow which I guess they do) to reuse the same instance, similarly to how we reuse instance in `dandi-cli`
- development: `./up develop` brings up all needed components (celery, rabbitmq etc) but uses flask from virtualenv/local installation and the code of this repository. `DATALAD_REGISTRY_INSTANCE_PATH` env var should be respected 
- deployment: thinking about it -- don't know yet what would be an ideal solution.  Let's just get `./up develop` mode going for now and then depending on where/how it would be deployed - provide a corresponding `./up deploy` or `./up build-deployment-image && ./up start-deploy`

jwodder:
Closes #42.


yarikoptic:
note: for some reason celery jobs aren't kicking in for me ATM (well -- with yesterday's version, started with `./up`) and don't see anything in `docker log` for those instances on that... yet to figure out how to troubleshoot and/or compare to behavior in master, where I believe it worked

jwodder:
@yarikoptic Try again with the latest code.  Using that, the celery containers come up fine for me.

yarikoptic:
This looks great, and seems to work for me nicely, thank you @jwodder!    Let's proceed!

Issue #25, opened 2021-03-23
registry-submit(-url?)


yarikoptic:
We need to provide easy means to register datasets we run into in a wild without requiring manual cloning them locally first.  Not sure if it is worth anyhow tuning `registry-submit` which is oriented more toward operations on locally present dataset, or create a new command e.g. `registry-submit-url [--stdin] [URL ...]`

- I think it would be valuable to provide interface to support multiple URLs at once to avoid re-creating tokens for separate invocations
- `--stdin` could then become particularly handy to just pipe them in without doing dances with `xargs` and limiting by number of URLs to be provided at once

It should be the server's job to receive a request to "process" a URL which would 
- if URL is not yet registered - clone and discover UUID and whatever else it could/should discover (yet to see if state of heads etc is associated with a URL)
- if that dataset (UUID) is found to be already known, add that extra url record
- if that dataset is not known, register it

above operations should be done out of band (possibly related: #24) and client (`registry-submit-url`) should just receive acknowledgement that "request is received"

yarikoptic:
for this we would need a new end-point, since current `/datasets/{ds_id}/urls/{url_encoded}` (used in https://github.com/datalad/datalad-registry/blob/master/datalad_registry_client/submit.py#L50) requires knowing uuid.  

So it might be worth adding top level `/urls` and `/urls/{url_encoded}` to parallel the `/datasets/` with the same "semantic" on GET vs PATCH, etc.

Then we could "trigger" datasets discovery/registration even from https://github.com/datalad/datalad-usage-dashboard

yarikoptic:
@jwodder let's start work on this from introducing the end-point/adding jobs to the queue, and we will decide on the command (either a new one or just re-use or refactor `registry-submit`)

jwodder:
@yarikoptic For `GET`, if the URL is not known, should the endpoint still clone the dataset in order to get its ID?  That would require blocking for what could be quite a while.

yarikoptic:
That is the goal - to clone and figure out id. Should be done out of band by a celery job, which already is there but for cases where id is known.

jwodder:
@yarikoptic How is a `GET` request supposed to return anything if the ID is looked up out-of-band?

yarikoptic:
Sorry, you asked about GET. For that one (not PATCH) as it does currently, I believe it just returns that is is unknown and doesn't trigger a job.

jwodder:
[...]

PR #45, opened 2021-08-30
Endpoint for URL without ID

jwodder:
Part of #25.

### DIFF
--- a/datalad_registry/datasets.py
+++ b/datalad_registry/datasets.py
@@ -50,7 +50,7 @@ def urls(url_encoded: str) -> Any:
 
     if request.method == "GET":
         lgr.info("Checking status of registering %s as URL", url)
-        resp: Any = {"url": url}
+        resp: dict = {"url": url}
         if row_known is None:
             status = "unknown"
         else:
--- a/datalad_registry/datasets.py
+++ b/datalad_registry/datasets.py
@@ -66,7 +66,7 @@ def urls(url_encoded: str) -> Any:
 
     elif request.method == "PATCH":
         if row_known is None:
-            tasks.collect_dataset_uuid.delay([url])
+            tasks.collect_dataset_uuid.delay(url)
         else:
             result.update({"update_announced": 1})
             db.session.commit()
--- a/datalad_registry/datasets.py
+++ b/datalad_registry/datasets.py
@@ -1,15 +1,19 @@
-"""Blueprint for /datasets view
+"""Blueprint for root view
 """
 
 import logging
+from typing import Any
 
 from flask import Blueprint
 from flask import jsonify
 from flask import request
 from flask import url_for
 
+from datalad_registry import tasks
 from datalad_registry.models import db
 from datalad_registry.models import URL
+from datalad_registry.utils import InvalidURL
+from datalad_registry.utils import url_decode
 
 lgr = logging.getLogger(__name__)
 bp = Blueprint("datasets", __name__, url_prefix="/v1/")
@@ -32,3 +36,38 @@ def datasets():  # No type hints due to mypy#7187.
         return jsonify({"next": pg_n,
                         "previous": pg_p,
                         "ds_ids": [i.ds_id for i in r.items]})
+
+
+@bp.route("urls/<string:url_encoded>", methods=["GET", "PATCH"])
+def urls(url_encoded: str) -> Any:
+    try:
+        url = url_decode(url_encoded)
+    except InvalidURL:
+        return jsonify(message="Invalid encoded URL"), 400
+
+    result = db.session.query(URL).filter_by(url=url)
+    row_known = result.first()
+
+    if request.method == "GET":
+        lgr.info("Checking status of registering %s as URL", url)
+        resp: Any = {"url": url}
+        if row_known is None:
+            status = "unknown"
+        else:
+            status = "known"
+            resp["ds_id"] = row_known.ds_id
+            resp["info"] = {
+                col: getattr(row_known, col, None)
+                for col in ["annex_uuid", "annex_key_count",
+                            "head", "head_describe"]}
+        lgr.debug("Status for %s: %s", url, status)
+        resp["status"] = status
+        return jsonify(resp)
+
+    elif request.method == "PATCH":
+        if row_known is None:
+            tasks.collect_dataset_uuid.delay([url])
+        else:
+            result.update({"update_announced": 1})
+            db.session.commit()
+        return "", 202
--- a/datalad_registry/tasks.py
+++ b/datalad_registry/tasks.py
@@ -98,33 +98,27 @@ def _extract_annex_info(repo) -> InfoType:
 
 
 @celery.task
-def collect_dataset_uuid(urls: Optional[List[str]] = None) -> None:
+def collect_dataset_uuid(url: str) -> None:
     from flask import current_app
 
     cache_dir = Path(current_app.config["DATALAD_REGISTRY_DATASET_CACHE"])
     cache_dir.mkdir(parents=True, exist_ok=True)
 
-    if not urls:
-        lgr.debug("Did not find URLs that needed UUIDs collected")
-        return
+    lgr.info("Collecting UUIDs for URL %s", url)
 
-    lgr.info("Collecting UUIDs for %s URLs", len(urls))
-    lgr.debug("URLs: %s", urls)
-
-    for url in urls:
-        result = db.session.query(URL).filter_by(url=url)
-        if result.first() is not None:
-            result.update({"update_announced": 1})
-        else:
-            ds_path = cache_dir / "UNKNOWN" / url_encode(url)
-            ds = clone_dataset(url, ds_path)
-            ds_id = ds.id
-            info = get_info(ds.repo)
-            db.session.add(URL(url=url, ds_id=ds_id, **info))
-            try:
-                ds_path.rename(cache_dir / ds_id[:3] / url_encode(url))
-            except OSError:
-                rmtree(ds_path)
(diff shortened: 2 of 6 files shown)


datalad_registry/datasets.py
```diff
@@ -32,3 +36,38 @@ def datasets():  # No type hints due to mypy#7187.
...
+        url = url_decode(url_encoded)
+    except InvalidURL:
+        return jsonify(message="Invalid encoded URL"), 400
+
+    result = db.session.query(URL).filter_by(url=url)
+    row_known = result.first()
+
+    if request.method == "GET":
+        lgr.info("Checking status of registering %s as URL", url)
+        resp: Any = {"url": url}
```
yarikoptic, line 53: why not 
```suggestion
        resp: dict = {"url": url}
```
?

datalad_registry/tasks.py
```diff
@@ -16,6 +18,31 @@
...
+    if ds_path.exists():
+        ds = dl.Dataset(ds_path)
+        ds_repo = ds.repo
+        ds_repo.fetch(all_=True)
+        ds_repo.call_git(["reset", "--hard", "refs/remotes/origin/HEAD"])
+    else:
+        ds = dl.clone(url, ds_path_str)
+    return ds
+
+def get_info(ds_repo: Any) -> InfoType:
```
yarikoptic, line 38:
```suggestion
def get_info(ds_repo: "GitRepo") -> InfoType:
```
? or if class must be loaded, then add `from datalad.support.gitrepo import GitRepo` among imports

datalad_registry/tasks.py
```diff
@@ -70,6 +97,37 @@ def _extract_annex_info(repo) -> InfoType:
...
+    if not urls:
+        lgr.debug("Did not find URLs that needed UUIDs collected")
+        return
+
+    lgr.info("Collecting UUIDs for %s URLs", len(urls))
+    lgr.debug("URLs: %s", urls)
+
+    for url in urls:
+        result = db.session.query(URL).filter_by(url=url)
+        if result.first() is not None:
```
yarikoptic, line 116: I know that this is how it is used elsewhere, but I wondered if there is some `.only` or `.first_only` which would otherwise error out if query returns multiple? (which shouldn't happen, so it is just to ensure to "error out than to chase the ghosts")

`url = db.Column(db.Text, primary_key=True)` and I do not see any other key, so I guess it ensures that only one record should be there.

datalad_registry/tasks.py
```diff
@@ -70,6 +97,37 @@ def _extract_annex_info(repo) -> InfoType:
     return info
 
 
+@celery.task
+def collect_dataset_uuid(urls: Optional[List[str]] = None) -> None:
```
yarikoptic, line 101: note: a list here is I guess to mimic the existing interface of `collect_dataset_info`  which takes a list of `ds_id, url` pairs... not sure yet if there is a benefit of "batching" multiple updates this way, and also makes name a bit misleading (should have been `collect_dataset_infos` or alike to signal that it is not a single one to be processed).

datalad_registry/tasks.py
```diff
@@ -70,6 +97,37 @@ def _extract_annex_info(repo) -> InfoType:
...
+            result.update({"update_announced": 1})
+        else:
+            ds_path = cache_dir / "UNKNOWN" / url_encode(url)
+            ds = clone_dataset(url, ds_path)
+            ds_id = ds.id
+            info = get_info(ds.repo)
+            db.session.add(URL(url=url, ds_id=ds_id, **info))
+            try:
+                ds_path.rename(cache_dir / ds_id[:3] / url_encode(url))
+            except OSError:
```
yarikoptic, line 126: Let's log detail of the error.
If exception was `OSError: [Errno 39] Directory not empty:` (can we access the errno? if so - use that), which means that we just have that clone already -- all good, log at `.debug`.  If it is some other error, log at `.warning` (or .error) since it is not expected.

datalad_registry/tasks.py
```diff
@@ -70,6 +97,37 @@ def _extract_annex_info(repo) -> InfoType:
...
+    lgr.debug("URLs: %s", urls)
+
+    for url in urls:
+        result = db.session.query(URL).filter_by(url=url)
+        if result.first() is not None:
+            result.update({"update_announced": 1})
+        else:
+            ds_path = cache_dir / "UNKNOWN" / url_encode(url)
+            ds = clone_dataset(url, ds_path)
+            ds_id = ds.id
```
yarikoptic, line 121: ds.id can be None if URL points to a regular git repo, which is without `.datalad/config` with the id. Then it would be None
```
$> python3 -c 'import datalad.api as dl; ds = dl.clone("https://github.com/datalad/datalad-registry", "/tmp/datalad-registry-clone"); print(ds.id)'
None 
```
In such case, we can still get info, but we should skip renaming and may be allow for  ds_id to be `None` in the model?  I guess change models to

```
ds_id = db.Column(db.Text, nullable=True)
```

?  Later PATCH could trigger re-checking of the repo and possibly assign UUID.

yarikoptic (changes requested):
Initial review, minor comments with the "major" one was to be able to "handle" pure repositories which do not yet have an `id`.

datalad_registry/datasets.py
jwodder: I was just copying the code from [here](https://github.com/datalad/datalad-registry/blob/7669bf4bd6597685119747e1b3025ee14d1f7297/datalad_registry/dataset_urls.py#L53).

datalad_registry/tasks.py
jwodder: That would be the [`scalar_one()`](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.scalar_one method.

datalad_registry/tasks.py
jwodder: Unbatched task.

datalad_registry/tasks.py
jwodder: Nevermind; it fails when testing with `--devserver`.  Quoting `"GitRepo"` works, though it'll cause problems if flake8 is added (which I hope to add alongside black once there are no outstanding PRs).

jwodder: Nevermind again; mypy doesn't like it when `"GitRepo"` quoted without importing the name (and I'm now going to add type checking to tox.ini).

datalad_registry/tasks.py
jwodder: Done.

datalad_registry/datasets.py
```diff
@@ -1,15 +1,19 @@
-"""Blueprint for /datasets view
+"""Blueprint for root view
 """
```
yarikoptic, line 2: wow github gives me an odd diff here

image

datalad_registry/datasets.py
yarikoptic: well, sooner or later we better tighten it up here and there

datalad_registry/datasets.py
jwodder: I see that too.  No idea why.

datalad_registry/datasets.py
jwodder: Fixed.

datalad_registry/tasks.py
yarikoptic: let's start using it then wherever it or `scalar_one_or_none` in this case I guess

jwodder: Done.

jwodder: It turns out that `scalar_one()` and `scalar_one_or_none()` are methods of the `Result` type but not of the `Query` type used by the ORM.  Hence, we need to stick with `first()`.

datalad_registry/tasks.py
yarikoptic: > Nevermind; it fails when testing with `--devserver`

heh heh -- asyncio strikes back (according to that original comment).  Should not be an issue after we release datalad 0.15.0, but I guess no need to worry about that ATM.

datalad_registry/tasks.py
```diff
@@ -112,26 +174,13 @@ def collect_dataset_info(
...
-            ds = dl.clone(url, ds_path_str)
-            ds_repo = ds.repo
-        info: InfoType = _extract_git_info(ds_repo)
-        info.update(_extract_annex_info(ds_repo))
+        abbrev_id = "None" if ds_id is None else ds_id[:3]
+        ds_path = cache_dir / abbrev_id / url_encode(url)
+        ds = clone_dataset(url, ds_path)
+        info = get_info(ds.repo)
+        if ds_id is None:
+            info["ds_id"] = ds.id
```
yarikoptic, line 182: Unlikely but ds_id can change... I think we should eventually handle those cases (rename directory etc), but for now I guess it could suffice to alert that we ran into such one
```suggestion
        if ds_id is None:
            info["ds_id"] = ds.id
        elif ds_id != ds.id:
            lgr.warning("A dataset with an ID (%s) got a new one (%s)", ds_id, ds.id)
```

yarikoptic (approved):
Thank you! Left one wishlist logging suggestion, and then could be merged to "tested real-life".

datalad_registry/tasks.py
```diff
@@ -112,26 +174,15 @@ def collect_dataset_info(
...
```
yarikoptic, line 182: just now noted: this line is reported not covered and it is the entire point of this PR AFAIK... odd... but let's proceed

yarikoptic:
Thank you @jwodder!  let's proceed although it seems that actual test awaiting for ds_id to be assigned is yet todo

PR #46, opened 2021-09-01
Support invoking `registry-submit` with neither `--sibling` nor `--url`

### LINKED ISSUE #32: registry-submit [--sibling|-s NAME] without URL
as an alternative to "manually" looping through `--sibling NAME`

jwodder:
Closes #32.

### DIFF
--- a/datalad_registry_client/opts.py
+++ b/datalad_registry_client/opts.py
@@ -67,23 +67,23 @@ def process_args(
         if sibling not in remotes:
             raise ValueError("Unknown sibling: {}".format(sibling))
 
-    urls: List[Tuple[str, str]] = []
+    urls: List[str] = []
     if url:
-        urls.append((url, url_encode(url)))
+        urls.append(url)
     elif sibling:
         url = repo.config.get("remote.{}.url".format(sibling))
         if url is None:
             raise ValueError("Could not find URL for {}".format(sibling))
-        urls.append((url, url_encode(url)))
+        urls.append(url)
     else:
         for r in repo.get_remotes(with_urls_only=True):
             url = repo.get_remote_url(r)
             assert url is not None
-            urls.append((url, url_encode(url)))
+            urls.append(url)
 
     endpoint = endpoint or repo.config.get(
         "datalad_registry.endpoint",
         DEFAULT_ENDPOINT)
     return dict(ds=ds, ds_id=ds_id,
-                sibling=sibling, urls=urls,
+                sibling=sibling, urls=[(url, url_encode(url)) for url in urls],
                 endpoint=endpoint)
--- a/datalad_registry_client/opts.py
+++ b/datalad_registry_client/opts.py
@@ -3,7 +3,9 @@
 
 from typing import Any
 from typing import Dict
+from typing import List
 from typing import Optional
+from typing import Tuple
 
 from datalad.distribution.dataset import Dataset
 from datalad.distribution.dataset import EnsureDataset
@@ -65,17 +67,23 @@ def process_args(
         if sibling not in remotes:
             raise ValueError("Unknown sibling: {}".format(sibling))
 
-    if not url:
-        if not sibling:
-            raise ValueError(
-                "Must specify URL to use when sibling isn't given")
+    urls: List[Tuple[str, str]] = []
+    if url:
+        urls.append((url, url_encode(url)))
+    elif sibling:
         url = repo.config.get("remote.{}.url".format(sibling))
-        if not url:
+        if url is None:
             raise ValueError("Could not find URL for {}".format(sibling))
+        urls.append((url, url_encode(url)))
+    else:
+        for r in repo.get_remotes():
+            u = repo.get_remote_url(r)
+            assert u is not None
+            urls.append((u, url_encode(u)))
 
     endpoint = endpoint or repo.config.get(
         "datalad_registry.endpoint",
         DEFAULT_ENDPOINT)
     return dict(ds=ds, ds_id=ds_id,
-                sibling=sibling, url=url, url_encoded=url_encode(url),
+                sibling=sibling, urls=urls,
                 endpoint=endpoint)
--- a/datalad_registry_client/opts.py
+++ b/datalad_registry_client/opts.py
@@ -76,7 +76,7 @@ def process_args(
             raise ValueError("Could not find URL for {}".format(sibling))
         urls.append((url, url_encode(url)))
     else:
-        for r in repo.get_remotes():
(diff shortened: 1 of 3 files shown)


datalad_registry_client/opts.py
```diff
@@ -64,17 +66,23 @@ def process_args(
...
+        urls.append((url, url_encode(url)))
+    elif sibling:
         url = repo.config.get("remote.{}.url".format(sibling))
-        if not url:
+        if url is None:
             raise ValueError("Could not find URL for {}".format(sibling))
+        urls.append((url, url_encode(url)))
+    else:
+        for r in repo.get_remotes():
+            u = repo.get_remote_url(r)
```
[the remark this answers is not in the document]
jwodder: @yarikoptic Did you mean to add `with_urls_only` to the `get_remotes()` call instead?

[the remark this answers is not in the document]
yarikoptic: Yes, sorry about that

datalad_registry_client/opts.py
```diff
@@ -64,17 +66,23 @@ def process_args(
...
+    if url:
+        urls.append((url, url_encode(url)))
+    elif sibling:
         url = repo.config.get("remote.{}.url".format(sibling))
-        if not url:
+        if url is None:
             raise ValueError("Could not find URL for {}".format(sibling))
+        urls.append((url, url_encode(url)))
+    else:
+        for r in repo.get_remotes():
```
yarikoptic, line 78:
```suggestion
        for r in repo.get_remotes(with_urls_only=True):
```

datalad_registry_client/opts.py
```diff
@@ -64,17 +66,23 @@ def process_args(
...
         url = repo.config.get("remote.{}.url".format(sibling))
-        if not url:
+        if url is None:
             raise ValueError("Could not find URL for {}".format(sibling))
+        urls.append((url, url_encode(url)))
+    else:
+        for r in repo.get_remotes(with_urls_only=True):
+            u = repo.get_remote_url(r)
+            assert u is not None
+            urls.append((u, url_encode(u)))
```
yarikoptic, line 81:
```suggestion
            url = repo.get_remote_url(r)
            assert url is not None
            urls.append((url, url_encode(url)))
```
for consistency with previous block where it is a `url`.

datalad_registry_client/opts.py
```diff
@@ -64,17 +66,23 @@ def process_args(
         if sibling not in remotes:
             raise ValueError("Unknown sibling: {}".format(sibling))
 
-    if not url:
-        if not sibling:
-            raise ValueError(
-                "Must specify URL to use when sibling isn't given")
+    urls: List[Tuple[str, str]] = []
+    if url:
+        urls.append((url, url_encode(url)))
```
yarikoptic, line 71: Why bother `url_encode`ing in this function (in 3 places) - why not to return just a list of `urls` and invoke `url_encode` in one spot whenever used outside and encoded value is needed?

yarikoptic (changes requested):
minor, possibly could be "disputed" ;)

datalad_registry_client/opts.py
jwodder: Done.

datalad_registry_client/opts.py
```diff
@@ -64,17 +66,23 @@ def process_args(
...
+            url = repo.get_remote_url(r)
+            assert url is not None
+            urls.append(url)
 
     endpoint = endpoint or repo.config.get(
         "datalad_registry.endpoint",
         "http://127.0.0.1:5000/v1")
     return dict(ds=ds, ds_id=ds_id,
-                sibling=sibling, url=url, url_encoded=url_encode(url),
+                sibling=sibling, urls=[(url, url_encode(url)) for url in urls],
```
yarikoptic, line 87: but why to do it `url_encode(url)` in this function at all and not outside whenever `url_encode`ing is actually needed, like [here](https://github.com/datalad/datalad-registry/pull/46/files#diff-e8ba59201fb7b17dd59ceb249e21c397fe879dce205a093ff03df88e13921fb8R51)?  outside code might decide to handle those urls differently so better that special handling is done where/when needed.
sorry if I am missing something obvious

datalad_registry_client/submit.py
```diff
@@ -38,36 +38,36 @@ def __call__(
...
-                                     timeout=1)
-            r_patch.raise_for_status()
-        except requests.exceptions.RequestException as exc:
-            yield dict(res_base, status="error",
-                       message=("Submitting URL failed: %s", exc))
-            return
-        yield dict(res_base, status="ok",
-                   message=("%s: %s", msg, url))
+            try:
+                r_patch = requests.patch(f"{base_url}/{ds_id}/urls/{url_encoded}",
```
yarikoptic, line 65: in my other comment I meant, why not just
```suggestion
                r_patch = requests.patch(f"{base_url}/{ds_id}/urls/{url_encode(url)}",
```
here and have `urls` to be just a list of urls

datalad_registry_client/opts.py
jwodder: That's just the way the code was when I found it.  However, encoding the URL in `opts.process_args()` does lead to the encoded URL being included in the result dict, which seems undesirable, so I'll change it.

jwodder: @yarikoptic On a related note, should the result dicts include everything that `opts.process_args()` returns?  Including `urls` at the very least seems odd, as it would probably be a better idea for each individual result dict to have a single `url` key for respective URL, and then there's the `ds_id` key....

yarikoptic: I agree -- a separate record per url would be better, since could have different status etc.  I would also add `url_encoded` as well in that result entry though to simplify possible troubleshooting etc.

datalad_registry_client/tests/test_submit.py
```diff
@@ -93,7 +120,7 @@ def test_submit_explicit_endpoint(tmp_path):
     assert_in_results(
         ds.registry_submit(url=ds.path, endpoint="abc", on_failure="ignore"),
         action="registry-submit", type="dataset",
-        path=ds.path, status="error")
+        path=ds.path, url=ds.path, status="error")
```
yarikoptic, line 123: note to myself: ATM allows for URL to just be a local path.  Could be a "feature" I guess for some local deployments to keep a "catalog" of available (locally) datasets.

yarikoptic:
Dear @jwodder , please resolve minor conflict in imports to your liking

jwodder:
@yarikoptic Conflicts resolved.

yarikoptic:
Thank you!

PR #47, opened 2021-09-03
Add registry-submit-urls command

jwodder:
Part of #25.

### DIFF
--- a/datalad_registry_client/submit_urls.py
+++ b/datalad_registry_client/submit_urls.py
@@ -1,0 +1,93 @@
+import logging
+from typing import Any
+from typing import Dict
+from typing import Optional
+from typing import Iterator
+from typing import List
+
+import requests
+from datalad import cfg
+from datalad.distribution.dataset import Dataset
+from datalad.interface.base import build_doc
+from datalad.interface.base import Interface
+from datalad.interface.results import get_status_dict
+from datalad.interface.utils import eval_results
+from datalad.support.constraints import EnsureNone
+from datalad.support.constraints import EnsureStr
+from datalad.support.param import Parameter
+
+from datalad_registry.utils import url_encode
+
+lgr = logging.getLogger("datalad.registry.submit_urls")
+
+
+@build_doc
+class RegistrySubmitURLs(Interface):
+    """Submit one or more URLs to a DataLad Registry instance."""
+
+    _params_ = {
+        "endpoint": Parameter(
+            args=("--endpoint",),
+            metavar="URL",
+            doc="""DataLad Registry instance to use (no trailing slash).
+            This defaults to the datalad_registry.endpoint option, if set,
+            or http://127.0.0.1:5000/v1 otherwise.""",
+            constraints=EnsureStr() | EnsureNone(),
+        ),
+        "urls": Parameter(
+            args=("urls",),
+            metavar="URL",
+            doc="""URLs to register""",
+            nargs="+",
+            constraints=EnsureStr() | EnsureNone(),
+        ),
+    }
+
+    @staticmethod
+    @eval_results
+    def __call__(
+        urls: List[str], endpoint: Optional[str] = None
+    ) -> Iterator[Dict[str, Any]]:
+        if endpoint is None:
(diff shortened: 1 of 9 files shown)


datalad_registry_client/submit_urls.py
```diff
@@ -0,0 +1,93 @@
...
+    """Submit one or more URLs to a DataLad Registry instance."""
+
+    _params_ = {
+        "endpoint": Parameter(
+            args=("--endpoint",),
+            metavar="URL",
+            doc="""DataLad Registry instance to use (no trailing slash).
+            This defaults to the datalad_registry.endpoint option, if set,
+            or http://127.0.0.1:5000/v1 otherwise.""",
+            constraints=EnsureStr() | EnsureNone(),
```
yarikoptic, line 35: I thought to suggest 
```suggestion
            This defaults to the datalad_registry.endpoint option.""",
            constraints=EnsureStr(),
            default=cfg.get("datalad_registry.endpoint", "http://127.0.0.1:5000/v1"),
```
and then would avoid the need for the code in the body of the function, but I do not see us using `cfg.get` in datalad itself in such cases, so may be there is a reason, and thus let's leave as is.

datalad_registry_client/submit_urls.py
```diff
@@ -0,0 +1,93 @@
...
+                    r = s.patch(f"{endpoint}/urls/{url_encoded}", timeout=1)
+                    r.raise_for_status()
+                except requests.exceptions.RequestException as exc:
+                    yield {
+                        **res_base,
+                        "url": url,
+                        "status": "error",
+                        "message": ("Submitting URL failed: %s", exc),
+                    }
+                    continue
```
yarikoptic, line 87: please add a test case submitting a URL to some unlikely to be running end point, e.g. `ENDPOINT + 'wrongendpoint/'` and getting that error

datalad_registry_client/submit_urls.py
yarikoptic: note: I also feel that may be later  should in `datalad_registry.__init__` just add a corresponding entry to https://github.com/datalad/datalad/blob/master/datalad/interface/common_cfg.py#L41 thus providing default there, and then use `cfg.obtain`, and thus avoiding need for default value any time we `get` it... do not remember that kitchen ATM to tell to do that for certain

yarikoptic (changes requested):
looks great.  yet to try. Left minor wishlist (test, const) and some notes.

datalad_registry_client/submit_urls.py
```diff
@@ -0,0 +1,93 @@
...
+class RegistrySubmitURLs(Interface):
+    """Submit one or more URLs to a DataLad Registry instance."""
+
+    _params_ = {
+        "endpoint": Parameter(
+            args=("--endpoint",),
+            metavar="URL",
+            doc="""DataLad Registry instance to use (no trailing slash).
+            This defaults to the datalad_registry.endpoint option, if set,
+            or http://127.0.0.1:5000/v1 otherwise.""",
```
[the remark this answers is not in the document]
jwodder: @yarikoptic Did you mean `datalad_registry_client.consts`?  `datalad_registry` doesn't use that string.

datalad_registry_client/submit_urls.py
jwodder: That's what [this test](https://github.com/datalad/datalad-registry/blob/4fc0275f4a6dd12c2184996441beb36d8e45d21a/datalad_registry_client/tests/test_submit_urls.py#L62) does, but the code in question isn't covered because the bad endpoint causes a failure earlier at the GET step.

datalad_registry_client/submit_urls.py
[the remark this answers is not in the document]
yarikoptic: yes, correct, sorry for misspecification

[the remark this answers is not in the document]
jwodder: Done.

datalad_registry_client/submit_urls.py
yarikoptic: ok, if tricky - let's postpone striving to cover it.

yarikoptic:
thank you @jwodder -- let's proceed!

PR #50, opened 2021-09-08
Add & apply linters

jwodder:


docs/openapi.yml
```diff
@@ -79,7 +79,6 @@ paths:
                   "ds_id": "6eeadb86-84be-11e6-b24c-002590f97d84",
                   "urls": ["https://datasets.datalad.org/labs/haxby/raiders/.git"]
                 }
-      responses:
```
yarikoptic, line 82: woohoo -- even caught/fixed a bug here!

yarikoptic:
This is great, thank you @jwodder !

PR #52, opened 2021-09-08
Use versioningit

### LINKED ISSUE #51: versiongit it
I have pushed a few tags, so now we can add proper versioning  instead of hardcoded 
```shell
$> git grep __version
datalad_registry/__init__.py:__version__ = "0.0.0"
```

jwodder:
Closes #51.

### DIFF
--- a/datalad_registry/__init__.py
+++ b/datalad_registry/__init__.py
@@ -1,5 +1,12 @@
-__version__ = "0.0.0"
+import sys
 
 from celery import Celery
+
+if sys.version_info[:2] < (3, 8):
+    from importlib_metadata import version
+else:
+    from importlib.metadata import version
+
+__version__ = version("datalad-registry")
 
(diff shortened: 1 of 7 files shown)


datalad_registry/__init__.py
```diff
@@ -1,5 +1,12 @@
-__version__ = "0.0.0"
+import sys
 
 from celery import Celery
 
+if sys.version_info[:2] < (3, 8):
+    from importlib_metadata import version
+else:
+    from importlib.metadata import version
```
yarikoptic, line 8: just a note to myself: I thought to just make 3.8 to be minimal version , but since it is a datalad extension, we better stay aligned with versions of python datalad release supports. ATM it is still 3.6 (which EOLs this Dec)

yarikoptic:
Thank you @jwodder. Let's proceed

PR #53, opened 2021-09-08
Better handle submitting the same URL to multiple endpoints at once

### LINKED ISSUE #48: Handle datasets submitted via `/urls/{url_encoded}` and `/datasets/{uuid}/urls/{url_encoded}` one after the other
(This is a to-do item for later; I'm not going to try to fix this at the moment.)

As of commit 4fc0275, if a dataset is submitted via `/urls/{url_encoded}` (which causes a row with a null `ds_id` to be added to the database) and then, before the celery task completes, submitted again via `/datasets/{uuid}/urls/{url_encoded}`, then the latter endpoint will query the database for the given UUID and URL, find nothing (because the UUID in the row isn't set), and then try to add a row with the same URL to the database, leading to an error.

jwodder:
Closes #48.

### DIFF
--- a/datalad_registry/dataset_urls.py
+++ b/datalad_registry/dataset_urls.py
@@ -36,17 +36,26 @@ def url(ds_id: str, url_encoded: str) -> Any:
     except InvalidURL:
         return jsonify(message="Invalid encoded URL"), 400
 
-    result = db.session.query(URL).filter_by(url=url, ds_id=ds_id)
+    result = db.session.query(URL).filter_by(url=url)
     # even here could lead to sqlite locking issue, I guess if
     # collect_dataset_info is in progress:
     # https://github.com/datalad/datalad-registry/issues/34
     row_known = result.first()
+
+    # We can't replace this check by adding `ds_id=ds_id` to the filter query,
+    # as then entries without ds_id set would not be returned, as though they
+    # were unregistered, leading to PATCH requests trying to register the URLs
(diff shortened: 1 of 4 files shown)


datalad_registry/dataset_urls.py
```diff
@@ -36,17 +36,26 @@ def url(ds_id: str, url_encoded: str) -> Any:
...
     # collect_dataset_info is in progress:
     # https://github.com/datalad/datalad-registry/issues/34
     row_known = result.first()
 
+    # We can't replace this check by adding `ds_id=ds_id` to the filter query,
+    # as then entries without ds_id set would not be returned, as though they
+    # were unregistered, leading to PATCH requests trying to register the URLs
+    # again by adding a row to the database, causing an error.
+    if row_known is not None and row_known.ds_id != ds_id:
+        return jsonify(message="UUID does not match value registered for URL"), 400
```
yarikoptic, line 50: I think it is ok for now, but let's add a comment to figure out for a better way to handle such use cases here:
```suggestion
        # TODO: figure out/provide logic/responses for datasets/urls which get their .id changed,
        # or yet to be processed (i.e. `not row_known.processed`)
        return jsonify(message="UUID does not match value registered for URL"), 400
```
?

jwodder: Comment added.

yarikoptic:
Thank you!

PR #56, opened 2021-09-08
Use an integer as the primary key

jwodder:
For improved performance.


yarikoptic:
good idea.  thank you!

PR #59, opened 2021-09-08
Give some columns better types

jwodder:


datalad_registry/overview.py
```diff
@@ -56,8 +55,8 @@ def overview():  # No type hints due to mypy#7187.
                 row[col] = getattr(info, col)
 
             ts = info.info_ts
-            if ts:
-                row["last_update"] = time.strftime("%Y-%m-%dT%H:%M%z", time.gmtime(ts))
+            if ts is not None:
```
yarikoptic, line 58: curious, why `if ts` isn't sufficient?

datalad_registry/overview.py
```diff
@@ -56,8 +55,8 @@ def overview():  # No type hints due to mypy#7187.
...
+                row["last_update"] = ts.strftime("%Y-%m-%dT%H:%M%z")
```
yarikoptic, line 59: while at it, let's add seconds?
```suggestion
                row["last_update"] = ts.strftime("%Y-%m-%dT%H:%M:%S%z")
```

yarikoptic:
note: merged other PRs, caused conflicts

datalad_registry/overview.py
jwodder: `if ts is None` avoids ambiguity about whether a timestamp of zero seconds past the epoch is true or false.  (I seem to recall there being an issue with the `datetime` class regarding this in the past, but looking at the docs now, the only relevant bit I can find is that `time` values at midnight were false prior to Python 3.5.)

jwodder:
@yarikoptic Conflicts resolved.

datalad_registry/overview.py
yarikoptic: good to know.  Thank you!

datalad_registry/overview.py
```diff
@@ -56,8 +55,8 @@ def overview():  # No type hints due to mypy#7187.
                 row[col] = getattr(info, col)
 
             ts = info.info_ts
-            if ts:
-                row["last_update"] = time.strftime("%Y-%m-%dT%H:%M%z", time.gmtime(ts))
+            if ts is not None:
+                row["last_update"] = ts.strftime("%Y-%m-%dT%H:%M:%S%z")
             else:
                 row["last_update"] = None
```
yarikoptic, line 61: unrelated to this PR, since not changing this line but since not code covered.  I would assume that it ( `db.Column`?) is nullable (by default), thus this should not blow up

jwodder: Yes, [`nullable` (generally) defaults to `True`](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.Column.params.nullable).

PR #61, opened 2021-10-13
Ignore warning about datalad.version deprecation

jwodder:


yarikoptic:
Could you please submit a PR to datalad maint fixing that?

If you could also add similar checking for deprecations to datalad - would be appreciated

yarikoptic:
> `-W error::DeprecationWarning:^datalad` to every invocation of `python -m nose`,

if you could check how well it would do it, I think it should be sufficient to just add to `.travis.yml`.  If there is too much and needs `-W` may be indeed we should abandon the idea (at least for now... may be eventually we would move to pytest there)

PR #62, opened 2022-10-04
Add codespell to lint

yarikoptic:
primary purpose of this PR is to assess current CI state while getting back to this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ DataLad registry -- work in progress
 ### Development setup
 
 Here are steps for setting up a development environment either 1)
-using a virtual environment for everything but the celery broker or 2)
+using a virtual environment for everything but the Celery broker, or 2)
 using containers for both the Flask and Celery components.  Note that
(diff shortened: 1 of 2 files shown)


yarikoptic:
I think it might be `test_url_bad_url_get` stalling -- happened to me locally and it was for a simple `get`, I was considering to use https://pypi.python.org/pypi/timeout-decorator but also tried to find some "built-in" way to timeout in flask test_client, then the fail ran away from me after I upgraded and run that test individually - then it just fails on 2nd argument of the `@pytest.mark.parametrize`.  so damn -- can't reproduce....

yarikoptic:
uff, today is not my day... I pushed 3bd34f337854d81ac2b2785477829bee001b10cf to this branch and it is there at https://github.com/datalad/datalad-registry/tree/enh-lint -- but PR doesn't update. grr

yarikoptic:
I believe this PR is good for what it was intended for . Unfortunately stalling tests (yet to figure out) forbid being sure but I think we better proceed

PR #66, opened 2022-11-01
Fix tests

### LINKED ISSUE #65: Tests started to stall
As could be seen from https://github.com/datalad/datalad-registry/actions/workflows/test.yml and I forgot any detail I discovered on where they stall when I quickly tried to figure it out.  But since didn't happen before it is something about dependencies I guess. @jwodder please try to figure it out

jwodder:
Closes #65.

There is currently still one test failure that I'm unsure how to fix.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.0.1
+    rev: v4.3.0
     hooks:
       - id: check-added-large-files
       - id: check-json
@@ -10,12 +10,12 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/psf/black
-    rev: 21.8b0
+    rev: 22.10.0
     hooks:
       - id: black
 
   - repo: https://github.com/PyCQA/isort
-    rev: 5.9.3
+    rev: 5.10.1
     hooks:
       - id: isort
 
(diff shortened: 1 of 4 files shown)


jwodder:
Current status: I can get the failing test to pass if I add the line `ses.expire_all()` above [this line](https://github.com/datalad/datalad-registry/blob/6003529f288d1622f95f3540ef32a81ed52d272a/datalad_registry/tests/test_tasks.py#L47), but I don't believe that should be necessary.  The test also passes if Flask-SQLAlchemy<3 is used instead of the latest version, or if `collect_dataset_info()` is changed from a Celery task to a regular function.  I am unable to create an MVCE that reproduces the test failure.

jwodder:
According to [this comment](https://github.com/pallets-eco/flask-sqlalchemy/issues/508#issuecomment-1250350587), under Flask-SQLAlchemy 3.0, each time a Flask app context is pushed, a new SQLAlchemy session is created, and it seems that updates to the database in one session are not reflected in the objects cached by another session.  Thus, when running tasks in tests, we need to avoid entering another app context (which happens due to [this code](https://github.com/datalad/datalad-registry/blob/cbf89dbdb07ddb0aab470ec3f6d15fdc372b1b3a/datalad_registry/factory.py#L39-L44)), and we can do that by invoking the tasks' `run()` methods instead of calling them like functions.

yarikoptic:
Thank you @jwodder for getting to the bottom of it!

PR #75, opened 2023-01-06
Extend metadata extracted from git-annex

candleindark:
Count and size of annexed files in working tree are now extracted and included in web UI.

### DIFF
--- a/datalad_registry/models.py
+++ b/datalad_registry/models.py
@@ -14,6 +14,8 @@ class URL(db.Model):  # type: ignore
     ds_id = db.Column(db.Text, nullable=True)
     annex_uuid = db.Column(db.Text)
     annex_key_count = db.Column(db.Integer)
+    annexed_files_in_wt_count = db.Column(db.Integer)
+    annexed_files_in_wt_size = db.Column(db.Text)
     info_ts = db.Column(db.DateTime(timezone=True))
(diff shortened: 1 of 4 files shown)


datalad_registry/models.py
```diff
@@ -14,6 +14,8 @@ class URL(db.Model):  # type: ignore
     ds_id = db.Column(db.Text, nullable=True)
     annex_uuid = db.Column(db.Text)
     annex_key_count = db.Column(db.Integer)
+    annexed_files_in_wt_count = db.Column(db.Integer)
+    annexed_files_in_wt_size = db.Column(db.Text)
```
yarikoptic, line 18: why is this a Text and not Integer?

candleindark: The return from calling `call_annex_records` on the repo has a str value for the `'size of annexed files in working tree'` key. Since the value is for display only, I didn't convert it to an `int`.

But, I guess it is a good idea to convert it to an `int` so that we can use it for searching the database, for example.

What do you think?

yarikoptic: More for sorting etc, and in general numbers should be numbers ;-)

candleindark: OK. I will make the change.

candleindark: The change is provided in https://github.com/datalad/datalad-registry/pull/77.

PR #77, opened 2023-01-07
Store size of annexed files as int

candleindark:
Store size of annexed files in working tree as an integer

### DIFF
--- a/datalad_registry/tasks.py
+++ b/datalad_registry/tasks.py
@@ -109,9 +109,9 @@ def _extract_annex_info(repo) -> InfoType:
         info['annexed_files_in_wt_count'] = working_tree_record[
             'annexed files in working tree'
         ]
-        info['annexed_files_in_wt_size'] = working_tree_record[
-            'size of annexed files in working tree'
-        ]
+        info['annexed_files_in_wt_size'] = int(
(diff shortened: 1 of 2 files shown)


datalad_registry/tasks.py
```diff
@@ -109,9 +109,9 @@ def _extract_annex_info(repo) -> InfoType:
         info['annexed_files_in_wt_count'] = working_tree_record[
             'annexed files in working tree'
         ]
-        info['annexed_files_in_wt_size'] = working_tree_record[
-            'size of annexed files in working tree'
-        ]
+        info['annexed_files_in_wt_size'] = int(
+            working_tree_record['size of annexed files in working tree']
+        )
```
yarikoptic, line 114: not to myself that for some reason (probably uniformity of the types of output regardless of `--bytes`) that `git annex info --bytes --json` does return those ints as strings:
```
❯ git -C datalad/openneuro/ds000001 annex info --bytes --json | jq . | grep size
  "local annex size": "0",
  "size of annexed files in working tree": "2415778654",
  "bloom filter size": "33554432 (0% full)",
```

yarikoptic:
thanks!

PR #78, opened 2023-01-13
Migrate code style to Black

candleindark:
Migrated code style to Black while preserving git blame history; Updated pre-commit repos; Add GitHub Actions workflow for Black;


tox.ini
```diff
@@ -59,8 +59,9 @@ source =
 doctests = True
 exclude = .*/,build/,dist/,instance/,venv/
 hang-closing = False
-max-doc-length = 80
-max-line-length = 80
+max-doc-length = 88
+max-line-length = 88
+extend-ignore = E203
```
asmacdo, line 64: Could you add a comment for E203? I can never remember my flake8 ignore codes

tox.ini
```diff
@@ -59,8 +59,9 @@ source =
...
 unused-arguments-ignore-stub-functions = True
 select = C,B,B902,B950,E,E242,F,I,U100,W
 ignore = B005,E203,E262,E266,E501,I201,W503
```
asmacdo, line 67: these also

tox.ini
candleindark: E203 is whitespace before ':'. It conflicts with the default behavior of Black per https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html.

I will add the comment.

tox.ini
candleindark: I actually don't know what those are either. I didn't put those in. I will find out what they are for when I reorganize the project gradually.

PR #79, opened 2023-01-17
Process multiple datasets in parallel

### LINKED ISSUE #38: collect_dataset_info - process datasets in parallel
code https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tasks.py#L90 ATM processes datasets serially in a `for` loop within a single celery job. IMHO we should process datasets in parallel (separate celery jobs? but some queue is needed I guess with ability to "debounce" multiple job requests for the same dataset).  ATM cannot be done really due to limitation of sqlite (#34).

candleindark:
Closes #38 and annotated the code with numerous todos.

### DIFF
--- a/datalad_registry/tests/test_tasks.py
+++ b/datalad_registry/tests/test_tasks.py
@@ -1,3 +1,5 @@
+import time
+
 import pytest
 
 from datalad_registry import tasks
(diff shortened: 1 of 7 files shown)


datalad_registry/tests/test_tasks.py
```diff
@@ -107,26 +114,34 @@ def test_collect_dataset_info_announced_update(app_instance, client, tmp_path):
...
         (ds.pathobj / "bar").write_text("bar")
         ds.save()
         repo.tag("v2", message="Version 2")
 
         url_encoded = url_encode(url)
         client.patch(f"/v1/datasets/{ds.id}/urls/{url_encoded}")
-        tasks.collect_dataset_info.run()
+
+        tasks.collect_dataset_info.delay()
+        time.sleep(1)  # Allow the task to run in the Celery worker
```
yarikoptic, line 136: is there a way to figure out that task was already ran so we could do something like (I have no clue in celery):

```python
task_handle = tasks.collect_dataset_info.delay()
elapsed = 0
while not task_handle.done():
   elapsed += 0.1
   if elapsed >= 5:
        raise AssertionError(f"Task  {task_handle} did not complete in time")
   time.sleep(0.1)
```

so we do not cause 1 second waits whenever it might finish quickly but also let it go on longer on slower systems/setups

candleindark: In the current setup of Celery in this project, without a backend, [there is no way to know the status of an invoked task through Celery](https://github.com/datalad/datalad-registry/issues/86).

I have removed all the `sleep` statements in the tests for we don't really need them in the current configuration of the tests, running in "eager mode". However, [this configuration with "eager mode" is itself a problem which needs to be corrected](https://github.com/datalad/datalad-registry/issues/87).

The solution for processing the datasets in parallel passes original intended tests. However, the test setup and the setup of Celery in the project need to be modified, and I have created https://github.com/datalad/datalad-registry/issues/87 and  https://github.com/datalad/datalad-registry/issues/86 respectively.

PR #80, opened 2023-01-18
Fix:  Solved the problem of cloned datasets not being stored in cache

candleindark:
By ensuring the existence of the containing directory of the destination of any cloned dataset, a clone dataset can now be store in the cache directory.


yarikoptic:
coolio, the mystery remains on how it worked for me, but let's proceed

candleindark:
> 

Indeed. I managed to see one dataset in the cache directory before this fix. One possibly explanation that the cache directory got populated before this fix is the `collect_dataset_info` function. It is executed periodically by the `celery` scheduler.

PR #81, opened 2023-01-19
 Ensure Celery tasks receive arguments in correct type

candleindark:
Utilize @validate_arguments from Pydantic to ensure arguments of Celery tasks are received in their intended types, not corrupted by the transmission in JSON.

This PR fixes the problem of sending a celery task with an argument of a tuple, but the responding celery worker receiving an argument of a list.

### DIFF
--- a/datalad_registry/tasks.py
+++ b/datalad_registry/tasks.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from shutil import rmtree
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-from pydantic import validate_arguments
+from pydantic import StrictStr, validate_arguments
 
 from datalad_registry import celery
 from datalad_registry.models import URL, db
@@ -164,7 +164,9 @@ def collect_dataset_uuid(url: str) -> None:
 
 @celery.task
 @validate_arguments
-def collect_dataset_info(datasets: Optional[List[Tuple[str, str]]] = None) -> None:
+def collect_dataset_info(
+    datasets: Optional[List[Tuple[StrictStr, StrictStr]]] = None
+) -> None:
     """Collect information about `datasets`.
 
     Parameters
--- a/datalad_registry/tasks.py
+++ b/datalad_registry/tasks.py
@@ -5,6 +5,8 @@ from pathlib import Path
 from shutil import rmtree
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+from pydantic import validate_arguments
+
 from datalad_registry import celery
 from datalad_registry.models import URL, db
 from datalad_registry.utils import url_encode
@@ -117,6 +119,7 @@ def _extract_annex_info(repo) -> InfoType:
 
 
 @celery.task
+@validate_arguments
 def collect_dataset_uuid(url: str) -> None:
     from flask import current_app
 
@@ -160,6 +163,7 @@ def collect_dataset_uuid(url: str) -> None:
 
 
 @celery.task
+@validate_arguments
 def collect_dataset_info(datasets: Optional[List[Tuple[str, str]]] = None) -> None:
     """Collect information about `datasets`.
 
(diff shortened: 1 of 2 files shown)


datalad_registry/tasks.py
```diff
@@ -117,6 +119,7 @@ def _extract_annex_info(repo) -> InfoType:
 
 
 @celery.task
+@validate_arguments
```
yarikoptic, line 122: why do we need `@validate_arguments` here? or just for general check/validation and not really "conversion" like in the case of tuple-vs-list in `collect_dataset_info`?

datalad_registry/tasks.py
```diff
@@ -160,6 +163,7 @@ def collect_dataset_uuid(url: str) -> None:
...
 def collect_dataset_info(datasets: Optional[List[Tuple[str, str]]] = None) -> None:
```
yarikoptic, line 167: for my own sake tested it out -- pydantic does quite a bit of casting, not yet 100% certain I like `int` -> `str` casting here:

```
❯ cat /tmp/testpydantic.py
from pydantic import validate_arguments
from typing import Optional, List, Tuple

@validate_arguments
def collect_dataset_info(datasets: Optional[List[Tuple[str, str]]] = None) -> None:
    return datasets


print(repr(collect_dataset_info([["list", ""]])))
print(repr(collect_dataset_info([("tuple", "")])))
print(repr(collect_dataset_info([("tuple with int not str", 1)])))
print(repr(collect_dataset_info([("tuple with 1 element", )])))

❯ python3 /tmp/testpydantic.py
[('list', '')]
[('tuple', '')]
[('tuple with int not str', '1')]
Traceback (most recent call last):
  File "/tmp/testpydantic.py", line 12, in <module>
    print(repr(collect_dataset_info([("tuple with 1 element", )])))
  File "/usr/lib/python3/dist-packages/pydantic/decorator.py", line 40, in wrapper_function
    return vd.call(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/pydantic/decorator.py", line 133, in call
    m = self.init_model_instance(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/pydantic/decorator.py", line 130, in init_model_instance
    return self.model(**values)
  File "/usr/lib/python3/dist-packages/pydantic/main.py", line 331, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for CollectDatasetInfo
datasets -> 0
  wrong tuple length 1, expected 2 (type=value_error.tuple.length; actual_length=1; expected_length=2)
```

how do you feel about such casting?

datalad_registry/tasks.py
candleindark: It does the conversion as well. However, for just converting list to tuple, we don't really need this. I chose this solution because it is terse and looks more uniform with other tasks that we may implement in the future that pass arguments of a Pydantic model type. (Passing arguments of a Pydantic model types to a Celery task is enabled by https://github.com/datalad/datalad-registry/pull/82)

datalad_registry/tasks.py
candleindark: This casting is a big part of what Pydantic do. In our case, it allows the sending a tuple through a JSON message to be received (after casting) as a tuple. I don't really need this feature for resolving the list to tuple problem as your solution demonstrated. However, using this feature make the code terse and more intuitive for it enables the caller to send a tuple as an argument and the callee the receive the argument as a tuple instead of a list.

This casting also make passing of complex data object as Pydantic models to Celery tasks possible, which is implemented in https://github.com/datalad/datalad-registry/pull/82.

I think FastAPI relies quite a bit on this nature of Pydantic as well.

With all that said, please let me know if you don't like this solution.

candleindark: If the [type casting](https://docs.pydantic.dev/usage/models/#data-conversion) is too wild for comfort, we can limit it by using [strict types](https://docs.pydantic.dev/usage/types/#strict-types) in Pydantic.

I will modify the solution using the strict types tomorrow.

candleindark: I have modified the solution. The wild type casting of primitive types such as casting from `int` to `str` will no longer happen.

yarikoptic:
Great, thanks!

PR #82, opened 2023-01-19
Feature: Add support of passing arguments of complex types to Celery tasks

candleindark:
By extending the JSON serializer used by Celery to support Pydanic models, we can pass arguments of complex types to Celery tasks. This should be a handy feature when we need to implement more evolved Celery tasks.

Please note the serializer used by the Celery app is still a JSON serializer and not a pickle one. Thus, we are not exposed to any of [the risks associated with a pickle serializer](https://docs.celeryq.dev/en/stable/userguide/security.html#serializers).


yarikoptic:
looks good , diff is a bit confusing like if #81 was not merged, let's proceed nevetherless

PR #83, opened 2023-01-20
Provide more precise dependency specification

candleindark:
Provide a more precise dependency specification by modifying the `setup.cfg` file and adding requirement files.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,6 +36,9 @@ tests =
     coverage>=7.0
     pytest>=7.0
     pytest-cov>=4.0
+
+dev =
+    %(tests)s
     tox>=3.0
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -33,9 +33,11 @@ datalad.extensions =
 
 [options.extras_require]
 tests =
-    coverage
-    pytest
-    pytest-cov
+    coverage>=7.0
+    pytest>=7.0
+    pytest-cov>=4.0
+    tox>=3.0
```
yarikoptic, line 39: What about doing smth like this so there is a dedicated `dev`:

```suggestion
[dev]
    %(tests)s
    tox>=3.0
```

PR #85, opened 2023-01-29
Restrict SQLAlchemy to version before 2.0

candleindark:
The latest SQLAlchemy, version 2.0, 01-26-2023, is currently incompatible with our project. Let's restrict the version of SQLAlchemy to before 2.0 for now.


yarikoptic:
fine with me

Issue #86, opened 2023-01-31
Configure the Celery service to have a backend


candleindark:
[The Celery service needs to have a backend service to communicate task statuses and results](https://stackoverflow.com/a/41546250/4635580). Set up a backend service for the Celery service and have the Celery service utilize it.

candleindark:
This is to be closed by #222.

candleindark:
#222 is complete and merged. Thus, this is done.

Issue #93, opened 2023-02-04
Failures in dataset population


candleindark:
Datalad registry instance failed to handle population of datasets listed on https://github.com/datalad/datalad-usage-dashboard intermittently.

The population of Datalad registry instance using the first 10 and first 20 active datasets on GitHub listed on https://github.com/datalad/datalad-usage-dashboard failed intermittently. Individual failures are unique in the sense that individual population attempt failed at gathering information about different datasets.

Resolving https://github.com/datalad/datalad-registry/issues/86 would be helpful in location the cause of this issue.

candleindark:
One observation is that the failure of one submission of urls to a Datalad registry instance using `registry_submit_urls` doesn't seem to affect the result of another submission of urls to the same instance using `registry_submit_urls`.

candleindark:
Another observation is that the failure of population of dataset info can fail for a subsequent submission of URLs using  `registry_submit_urls` to a Datalad registry instance as well as the first submission to the Datalad registry instance.

yarikoptic:
- what failures did you observe?  
- any structure to them or logs?

yarikoptic:
#134 and #135 might be of this kind but more specific.  #133 might provide  an instrumentation to ease debugging.
Do you want to keep this issue open @candleindark or should we consider more specific to be replacement to it?

candleindark:
I think we should keep this open for now. #134 and #135 might be manifestations of this problem but this problem can possibly appear in a different way.

PR #96, opened 2023-02-08
Add Celery backend

candleindark:
Configure Celery to have a backend. This PR resolves https://github.com/datalad/datalad-registry/issues/86.

### DIFF
--- a/datalad_registry/config.py
+++ b/datalad_registry/config.py
@@ -21,7 +21,7 @@ def _log_level() -> Union[str, int]:
 
 class Config(object):
     CELERY_BROKER_URL = os.environ.get("CELERY_BROKER_URL", "amqp://localhost:5672")
-    CELERY_result_backend = os.environ["CELERY_RESULT_BACKEND"]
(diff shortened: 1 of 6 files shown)


candleindark:
The current state of this PR should have added a backend to the Celery service correctly. However, the tests for the project need to be adjusted. The test GitHub Action may also need to be adjusted as well for testing now requires a running Redis service.

The current state, with a backend set up, should be useful in resolving https://github.com/datalad/datalad-registry/issues/93.

yarikoptic:
oh, so current run on typhon is working ok without celery backend?

candleindark:
> oh, so current run on typhon is working ok without celery backend?

The backend service is running in Typhon but not utilized.

The tests need to be updated before I can merge this code.

candleindark:
Frankly, I want to rewrite some of the code that set up the Flask app and Celery app. Those code currently are really messy. Rewriting those code will also re-attach the backend but possibly in a even more uniform way, along with other configurations.

yarikoptic:
do we still need this PR in the long run?

candleindark:
This solution has been incorporated into #222. Thus, this PR should be closed.

PR #97, opened 2023-02-08
Modify broker container specifications

candleindark:
This PR changes few specs for the Celery broker:

1. Set a password
2. For security, restrict access to localhost only
3. Expose only the relevant environment variables


yarikoptic:
if works, I don't see why not ;)

Issue #89, opened 2023-01-31
Run via rootless podman or singularity/apptainer?


yarikoptic:
We should check with @asmacdo if that would be possible to avoid possible "kill all by using docker" scenario ;)

asmacdo:
huh?

asmacdo:
Without fully understanding the use case I think the answer is yes.

With podman, we have 4 options.
- run podman as root, run container as root
- run podman as root, run container as non-root (specify -u someuser)
- run podman as non-root, run the container as root (ex. UID 0 inside the container maps to UID 1000)
- run podman as non-root, run the container as non-root. safety first!

yarikoptic:
FWIW we have `podman` installed on smaug now.
In development mode we need to start `db` and `broker` from https://github.com/datalad/datalad-registry/blob/master/docker-compose.yml . 
https://github.com/datalad/datalad-registry/blob/master/up is the script used for that purpose. 

Can you @asmacdo work with @candleindark and research into possibility of the "non-root" execution of those instances?  I think it should be feasible since really there is no need for "outside root" - ports they use are not privileged so for that root also would not be needed.

candleindark:
@yarikoptic Podman on Smaug is currently @ 3.0.1. Can we have it updated to the current version 4.4.1?

yarikoptic:
I've tried to backport 4.3.1 (present in debian unstable) but "no good".

Why would we need 4.4.1?  is there a way for user-space installation (attn @asmacdo)?

asmacdo:
Seems like it's **possible** but probably not worth the effort.

https://github.com/containers/podman/issues/3100

candleindark:
> I've tried to backport 4.3.1 (present in debian unstable) but "no good".
> 
> Why would we need 4.4.1? is there a way for user-space installation (attn @asmacdo)?

All the documentation I have been reading is at version 4+. I was just wondering if we can upgrade to version 4+ just for consistency. It actually don't have to be 4.4.1 in particular.

candleindark:
Datalad-registry has been run by Podman for more than 7 days. I think we should close this issue.

yarikoptic:
feel welcome to close issues you consider addressed. Where possible/relevant -- reference a commit or PR which closed it.  In this case I guess it might benefit from a quick section in README.md on podman and settings to use to avoid containers being killed.

candleindark:
Note:
[...]

PR #98, opened 2023-02-10
BF: Handle IncompleteResultsError in cloning dataset

candleindark:
Cloning a dataset sometimes can fail with an IncompleteResultsError. Simply retrying will resolve the problem in most cases.

This PR is only a step toward a solution for https://github.com/datalad/datalad-registry/issues/93. A complete solution would require the handling of the case in which all retries of cloning fail. However, with this PR, the population of datasets would be stable enough for the work for https://github.com/datalad/datalad-registry/issues/89 to start.

### DIFF
--- a/datalad_registry/tasks.py
+++ b/datalad_registry/tasks.py
@@ -3,9 +3,11 @@ import errno
 import logging
 from pathlib import Path
 from shutil import rmtree
+import time
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 from celery import group
+from datalad.support.exceptions import IncompleteResultsError
 from pydantic import StrictStr, validate_arguments
 
 from datalad_registry import celery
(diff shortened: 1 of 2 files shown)


datalad_registry/tasks.py
```diff
@@ -33,7 +34,22 @@ def clone_dataset(url: str, ds_path: Path) -> Any:
...
-        ds = dl.clone(url, ds_path_str)
+        max_incomplete_results_errs = 5
+        incomplete_results_err_count = 0
+
+        # Clone the dataset @ url into ds_path
+        while True:
+            try:
+                ds = dl.clone(url, ds_path_str)
+            except IncompleteResultsError:
+                lgr.warning(f"Incomplete results error in cloning {url}. Trying again.")
```
yarikoptic, line 45: `os.unlink(ds_path_str)` before retrying to ensure that follow up re-cloning is not somehow skipping since target exists but not fully kosher.

yarikoptic:
in the future checkout the tenacity library for generic framework for retries

PR #103, opened 2023-02-23
Use bind mounts instead of named volumes

candleindark:
Use bind mounts instead of named volumes for easy of allocation of resources in the host machine.


yarikoptic:
didn't try but if works -- great!

PR #104, opened 2023-02-23
Unique urls

candleindark:
Ensure the URLs submitted by each run of the `population.py` script are unique. This is useful in getting the actual number of datasets to be handled by the datalad registry instance.


yarikoptic:
looks good. thanks

PR #106, opened 2023-02-23
Use dict to preserve order of retrieved URLs for population

candleindark:
While [dictionaries are insertion ordered](https://stackoverflow.com/a/39980744/4635580), sets are not. It is better to use a dictionary to preserve the order of dataset URLs read from the usage dashboard so that we get the URLs for submission to the Datalad registry instance in a consistent order across multiple runs of the `population.py` script.


yarikoptic:
sounds good

PR #107, opened 2023-02-28
Use --always with git describe so we always have some

yarikoptic:


yarikoptic:
I think lint concern is unrelated right?

candleindark:
> 



> I think lint concern is unrelated right?

I can't really tell for I can't access the changes Black has put in. However, it should be fine to just merge. Black will enforce the format next time I modify the same file.

PR #115, opened 2023-03-01
HTTP Method handling

candleindark:
Remove unneeded checks of HTTP methods inside view functions, and make acceptable HTTP methods more explicit.


yarikoptic:
Makes sense , thanks!

PR #113, opened 2023-03-01
Establish model for meta data of URL

candleindark:
This pull request is intended to resolve https://github.com/datalad/datalad-registry/issues/108.

PR #118, opened 2023-03-05
Remove customized table name

candleindark:
[`__tablename__` doesn't need to be defined when using Flask-SQLAlchemy's model](https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/models/#defining-models). Let's remove the customized table name and adopt the naming convention implemented by Flask-SQLAlchemy.

### DIFF
--- a/datalad_registry/models.py
+++ b/datalad_registry/models.py
@@ -8,9 +8,6 @@ migrate = Migrate()
 
 
 class URL(db.Model):  # type: ignore
-
-    __tablename__ = "urls"
-
     id = db.Column(db.Integer, primary_key=True, nullable=False)
     url = db.Column(db.Text, nullable=False, unique=True)
     ds_id = db.Column(db.Text, nullable=True)
--- a/migrations/versions/769fc6a4d2ea_removing_customized___tablename___urls_.py
+++ b/migrations/versions/769fc6a4d2ea_removing_customized___tablename___urls_.py
@@ -1,0 +1,22 @@
+"""Removing customized __tablename__ urls, and let URL model to assume default name
+
+Revision ID: 769fc6a4d2ea
+Revises:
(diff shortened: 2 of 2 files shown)


yarikoptic:
wouldn't this need some kind of migration so we could just keep using already collected data in DB?

candleindark:
> wouldn't this need some kind of migration so we could just keep using already collected data in DB?

Yes. I believe so. I rather do it right though, and we need to do migration sooner or later. Even without migration, the repopulation of all the dataset we have takes about 2 hours on Typhon.

For migration, can we use https://github.com/miguelgrinberg/flask-migrate?

yarikoptic:
even though quick to repopulate - might be nice to practice migration.  that repo sounds good -- popular etc. See how it could be all "automated" nicely.

yarikoptic:
FWIW the only other flask app I use/know is https://salsa.debian.org/snapshot-team/snapshot  and it seems that one had ad-hoc logic implemented for each upgrade: https://salsa.debian.org/snapshot-team/snapshot/-/tree/master/db

@vmdocua did you ever have to deal with flask apps and their DB upgrades by any chance?

candleindark:
> 

No, I have not, but I will try it after finishing https://github.com/datalad/datalad-registry/pull/113.

candleindark:
This PR is dependent on https://github.com/datalad/datalad-registry/pull/126. Rebase this PR after https://github.com/datalad/datalad-registry/pull/126 is successfully merge to master to get the relevant commits.

yarikoptic (approved):
If that is what you want to do -- go ahead , merge away ;)

candleindark:
> If that is what you want to do -- go ahead , merge away ;)

Thanks for the freedom. I appreciate it.

PR #128, opened 2023-03-13
Add docker as a dependency

candleindark:
By adding this dependency, we can utilize [Docker Engine API](https://docs.docker.com/engine/api/) for testing and development in Python.


yarikoptic (changes requested):
"we can" or "we do"? I see no point of adding a dependency which we do not use yet . If there is a use of it you are planing to do as a tool, please add to CONTRIBUTING.md

candleindark:
I see your point. I will include this in a PR that actual uses it.

PR #137, opened 2023-03-14
Replace the use of `yield` in pytest fixtures

candleindark:
Replace the use of `yield` with `return` in pytest fixtures when there is nothing to be clean up afterward. The use of `yield` in `app_instance` is also problematic because it puts all tests using `app_instance` in the created app's app context implicitly and unnecessarily. For more detailed info, please checkout out https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/contexts/#tests.

### DIFF
--- a/datalad_registry/tests/conftest.py
+++ b/datalad_registry/tests/conftest.py
@@ -104,7 +104,7 @@ def _app_instance(dockerdb, tmp_path_factory, cache_dir):
 
     socket.setdefaulttimeout(2)  # seconds
     app = create_app(config)
-    yield AppInstance(app, db)
+    return AppInstance(app, db)
 
 
 @pytest.fixture
@@ -114,14 +114,14 @@ def app_instance(_app_instance):
     If you just need the client, you can use the `client` fixture
     instead.
 
-    Yields
+    Returns
     ------
     AppInstance namedtuple with app, db, and client fields.
     """
     with _app_instance.app.app_context():
         db.drop_all()
         db.create_all()
-        yield _app_instance
+        return _app_instance
 
 
 @pytest.fixture
@@ -131,4 +131,4 @@ def client(app_instance):
     If you need to access to the application or database, use the
     `app_instance` fixture instead.
     """
-    yield app_instance.app.test_client()
+    return app_instance.app.test_client()


candleindark:
> `pytest` recommends `yield` vs `return` in fixtures: https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended

I got the idea that `yield` was recommended over `return` when I initially encountered the documentation of https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended as well. However, if you read that piece of documentation in the greater context of https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization. I think `pytest` is recommending using `yield` fixtures over [adding finalizer directly](https://docs.pytest.org/en/6.2.x/fixture.html#adding-finalizers-directly) in leveraging the teardown system and not recommending the use of `yield` over `return` in fixtures in general.

> 
> > The use of `yield` in `app_instance` is also problematic because it puts all tests using `app_instance` in the created app's app context implicitly and unnecessarily. For more detailed info, please checkout out https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/contexts/#tests.
> 
> please digest it for me a bit more since there I only see example which uses `yield` one
> 
> ```python
> @pytest.mark.fixture
> def app_ctx(app):
>     with app.app_context():
>         yield
> ```
> 
> so it feels that our `app_instance` should remain using `yield`. 

According to [flask-sqlalchemy's documentation](https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/contexts/#tests), "Only push a context exactly where and for how long it’s needed for each test. Do not push an application context globally for every test, as that can interfere with how the session is cleaned up." We are doing pretty much the opposite in https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L121-L125 for we put all tests requiring the datalad_registry flask app instance, directly or indirectly, in the app's context regardless of necessity. Tests that require the  datalad_registry flask app instance don't necessarily need to be in the app's context.


Before this example

```python
import pytest

@pytest.mark.fixture
def app_ctx(app):
    with app.app_context():
        yield

@pytest.mark.usefixtures("app_ctx")
def test_user_model(app):
    user = User()
    db.session.add(user)
    db.session.commit()
```

is given, the [documentation](https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/contexts/#tests) states "if you find yourself writing many tests like that, you can use a pytest fixture to push a context for a specific test." So it's an example of how a flask app's context can be pushed for a specific test using a fixture, for convenience. It doesn't contradict with the idea that a flask app's context shouldn't be pushed unnecessarily for tests.

To make my point clear, I think I can achieve the same fix this PR achieves just by replacing

```python
    with _app_instance.app.app_context():
        db.drop_all()
        db.create_all()
        yield _app_instance
```

from [the original code](https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L121-L125) with

```python
    with _app_instance.app.app_context():
        db.drop_all()
        db.create_all()
    yield _app_instance
```

without changing any `yield` to `return`.

With that said there is an additional benefit of using `return` instead of `yield` in the fixtures involved in this PR. All the involved fixtures don't require cleanup after their invocation. Thus, using `return` instead of `yield` can signal to the reader of the code that the involved fixtures are of this nature.



>Is this PR solving any issue?

Yes, as stated in [flask-sqlalchemy's documentation](https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/contexts/#tests), putting tests in a flask app's application context unnecessarily "can interfere with how the session is cleaned up." I have seen this problem manifested in our project. It was previously circumvented in the manner depicted in [this commit](https://github.com/datalad/datalad-registry/commit/cbf89dbd), and it is currently circumvented by closing the `db.session` manually as in https://github.com/datalad/datalad-registry/blob/a20bb446bf2ac32499b3860b9f3ce6cc503cefe2/datalad_registry/tests/test_tasks.py#L90

candleindark:
With what is stated in the above comment, it may be a good idea to change the title of this PR to "Avoid putting tests in datalad_registry flask app's application context unnecessarily" because that is the essential goal of this PR.

In the same vein, it may be a good idea to just modify 

```python
    with _app_instance.app.app_context():
        db.drop_all()
        db.create_all()
        yield _app_instance
```

from [the original code](https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L121-L125) to

```python
    with _app_instance.app.app_context():
        db.drop_all()
        db.create_all()
    yield _app_instance
```

in this PR though the current fix has additional benefits.

@yarikoptic Let me know what you think.

yarikoptic:
but why is it unnecessarily if flask example itself does

```python
@pytest.mark.fixture
def app_ctx(app):
    with app.app_context():
        yield
```
ie yielding fixture while in the context?

well - tests still pass so I guess it might indeed be "unnecessarily" but I want to understand why you think it is so?

candleindark:
> but why is it unnecessarily if flask example itself does
> 
> ```python
> @pytest.mark.fixture
> def app_ctx(app):
>     with app.app_context():
>         yield
> ```
> 
> ie yielding fixture while in the context?
> 
> well - tests still pass so I guess it might indeed be "unnecessarily" but I want to understand why you think it is so?

The point of the given example is to make a fixture that would put a test using it in a flask app context. By yielding inside the app's context the example does exactly that.

The point of [the `app_instance` fixture](https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L110-L125) in our code, not what it is actually doing now, really is to provide a test using it the datalad_registry flask app and not to put the test in the app's context. However, by yielding inside the app's context, [the `app_instance` fixture](https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L110-L125) unnecessarily put any test using it inside the datalad_registry flask app's context as well. In other words, [the `app_instance` fixture](https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L110-L125) in our code is really the following two fixtures combined.

```python
@pytest.fixture
def lean_app_instance(_app_instance):
    with _app_instance.app.app_context():
        db.drop_all()
        db.create_all()
    yield _app_instance


@pytest.fixture
def app_instance_ctx(_app_instance):
    with _app_instance.app.app_context():
        yield
```

Using [the `app_instance` fixture](https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L110-L125) in our code in a test such as:

```python
def test_foo(app_instance):
    ...
    ...
```

is the the same as
```python
@pytest.mark.usefixtures("app_instance_ctx")
def test_foo(lean_app_instance):
    ...
    ...
```

Thus, [the `app_instance` fixture](https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L110-L125) in our code always  put a test that needs the datalad_registry flask app in the app's context.

Since the existing tests in our project that need the datalad_registry flask app do not need to be put in the app's context, i.e.:
```
def test_a_test_in_our_project(lean_app_instance):
    ...
    ...
```
using  [the `app_instance` fixture](https://github.com/datalad/datalad-registry/blob/master/datalad_registry/tests/conftest.py#L110-L125) in our code unnecessarily put these tests in the app's context and interfere with how the database session is cleaned up.

PR #138, opened 2023-03-20
Make sure error responses from /api/ are in JSON format

candleindark:
This PR is intended to resolve https://github.com/datalad/datalad-registry/issues/121.

This PR is built upon https://github.com/datalad/datalad-registry/pull/113. It should be rebased after https://github.com/datalad/datalad-registry/pull/113 is merged to master to get the relevant commit.

PR #139, opened 2023-03-21
Enable testing mode for the Datalad-registry Flask app in tests

candleindark:
This PR is based on https://github.com/datalad/datalad-registry/pull/138. It should be rebased after https://github.com/datalad/datalad-registry/pull/138 is merged to get the relevant commits of this PR.


yarikoptic:
what would be the effect of this?

Issue #132, opened 2023-03-14
sorting by numbers is rendered useless by empty (no) values


yarikoptic:
If I sort by any of those [Annex keys](http://localhost:5000/overview/?sort=keys-desc)	[Annexed files in working tree](http://localhost:5000/overview/?sort=annexed_files_in_wt_count-asc), instead of getting first the ones with the largest numbers I get first the ones without any.  I think it would be ok to sort them **after** the ones with the lowest (0) number.

PR #142, opened 2023-03-23
Sort with NULL values at the end of the result

candleindark:
This PR is intended to resolve https://github.com/datalad/datalad-registry/issues/132.

With the changes in this PR, all items with NULL values are allocated to the end of the result of a sort, regardless of the order of the sort, descending or ascending.

These changes have been made available in the running instance of Datalad-registry on Typhon

### DIFF
--- a/datalad_registry/overview.py
+++ b/datalad_registry/overview.py
@@ -4,6 +4,7 @@
 import logging
 
 from flask import Blueprint, render_template, request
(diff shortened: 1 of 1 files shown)


yarikoptic:
cool. FWIW -- sample page to see them there at the end http://localhost:5000/overview/?sort=annexed_files_in_wt_count-asc&page=149

Issue #133, opened 2023-03-14
Add and expose basic audit


yarikoptic:
While looking at http://localhost:5000/overview/?sort=annexed_files_in_wt_count-desc&page=5 I saw that e.g. https://github.com/OpenNeuroDerivatives/ds003242-fmriprep.git has no metadata for it, but cloning it locally showed that it has lots of annexed files in the tree etc.  So something did happen which prevented populating its record with metadata.  It would probably hard to impossible to figure out what has happened.

I think we might better right away establish basic audit for our entries.  Let's 

- create a table with model along the
  - id: int
  - url_id: int (external key from URL)
  - datetime: db.DateTime
  - success: bool
  - message: str
- for the tasks of clone, update, get basic info, extract metadata - record that it happened or errored out, storing error out detail in `message`
- add endpoint for `/audit` for a given url to just return json list of those records
- add column `audit` with count of audit messages and URL which would lead to that `/audit` endpoint (we might improve it later on, first just want to see all)

candleindark:
I think it is a good idea to add the audit feature.

For the particular example of https://github.com/OpenNeuroDerivatives/ds003242-fmriprep.git, I am able to see the the `metalad_core` metadata listed. Do you mean some other URL instead?

candleindark:
The monitoring service collects and logs information of each execution of a task in a database file. This information of each execution includes the following.
1. The task that is executed
2. The arguments used in the execution
3. The date and time the task is received by the worker
4. The date and time the task is started by the worker
5. The date and time the task is completed by the worker
6. The result of the execution or the exception resulted with traceback
7. If the task has been retried
8. How many times the task has been retried if retried has been issued

From the arguments used in the execution, we can find out exactly which dataset URL is involved in a task execution that involves a dataset URL. From a logged exception with traceback, we can find out exact how a task fails. The logging of time in various aspects of running a task in the Celery worker by the monitoring services is more than fulfilling the intents of the originally proposed audit system. All the logged information regarding executing/executed tasks are also available through the API of the monitoring service at [/api/tasks](https://flower.readthedocs.io/en/latest/api.html#get--api-tasks)

Because of the aforementioned properties of the monitoring service, I think the monitoring service fulfills the goal of this issue and this issue should be closed as complete.

Note: The monitoring service offers many other features beyond what is mentioned, such as monitoring Celery workers in the cluster and changing the properties of these workers. For more information, please check out its [official documentation](https://flower.readthedocs.io/en/latest/).

[...]

Issue #125, opened 2023-03-10
Reorganize local cache


candleindark:
Currently, the initial layer of the cache are directories based on UUID of datasets. Consider naming the directories using the first few letters of the hashed representation of URLs instead.

yarikoptic:
I also wonder if URL could be represented in some better form, still reflective of original URL, than base64?

ha -- we should adopt one used by git-annex itself for URL backend

```
❯ git annex addurl --fast https://github.com/datalad/datalad-registry/issues/125
addurl https://github.com/datalad/datalad-registry/issues/125 (to github.com_datalad_datalad-registry_issues_125) ok
(recording state in git...)
❯ ls -l
lrwxrwxrwx 1 yoh yoh 146 Mar 10 13:10 github.com_datalad_datalad-registry_issues_125 -> .git/annex/objects/50/Wg/URL--https&c%%github.com%datalad%datalad-registry%issues%125/URL--https&c%%github.com%datalad%datalad-registry%issues%125
```
so pretty much it is just URL encoding of the url itself!

```shell
In [13]: from urllib.parse import quote

In [14]: quote('https://github.com/datalad/datalad-registry/issues/125', safe='')
Out[14]: 'https%3A%2F%2Fgithub.com%2Fdatalad%2Fdatalad-registry%2Fissues%2F125'
```

OR even better -- why not to use use the

```
In [15]: quote('https://github.com/datalad/datalad-registry/issues/125')
Out[15]: 'https%3A//github.com/datalad/datalad-registry/issues/125'
```
 and without any leading folders based on checksum since we would get hierarchy naturally now unless we have thousands of subfolders under 1 single domain path, besides like for dandizarrs etc. But overall I don't think it is often to happen, so I would not even sweat about it.


Here is a list of all in usage dashboard

```shell
$ jq '.github[].url' < datalad-repos.json | tr '"' ' ' | python3 -c 'from urllib.parse import quote; import sys; print("\n".join(quote(l.strip()) for l in  sys.stdin.readlines()))'
```
available at http://www.oneukrainian.com/tmp/datalad-repos-urlquoted.txt .

I didn't check though either at the level current API we would not get a trouble using such URLs with `/` in them. If we do -- may be within API requests they would be needed to be "without safe /", so we would do `unquote/quote(safe="/")` to get to the local folder one.

candleindark:
I prefer to have the URLs to be encoded in a human readable form as well.

When you mentioned
```
[...]

PR #159, opened 2023-04-24
Processing of dataset URLs with the new cache organization

candleindark:
This PR is a new implementation of the processing of dataset URLs with [the new cache organization](https://github.com/datalad/datalad-registry/issues/125) and in support of [the V2 API](https://github.com/datalad/datalad-registry/pull/149). It lays the foundation for handling the race condition in accessing the cache, updating the dataset URLs information periodically, and [auditing Celery actions](https://github.com/datalad/datalad-registry/issues/133). Many of the extractions of information to populate a dataset URL object have been rewritten (re-organized in some cases) to handle edge conditions properly and/or to obtain results in more appropriate datatypes.


candleindark:
The remaining issues before this PR can be merged are

1. The tests for one reason or another are not using database connection economically.  A possible solution can be found https://stackoverflow.com/questions/66876181/how-do-i-close-a-flask-sqlalchemy-connection-that-i-used-in-a-thread
2. Some small problem with the linter or the setup of it.

candleindark:
> The remaining issues before this PR can be merged are
> 
> 1. The tests for one reason or another are not using database connection economically.  A possible solution can be found https://stackoverflow.com/questions/66876181/how-do-i-close-a-flask-sqlalchemy-connection-that-i-used-in-a-thread
> 2. Some small problem with the linter or the setup of it.

Thanks to @yarikoptic's hint during the weekly meeting, I took another look at how the `flask_app` fixture was setup. The Postgres connection issue is now resolved.

datalad_registry/models.py
```diff
@@ -25,6 +25,9 @@ class URL(db.Model):  # type: ignore
     #: Whether initial data has been collected for this URL
     processed = db.Column(db.Boolean, default=False, nullable=False)
 
+    # The path in the local cache where a copy of the dataset at the URL is stored
+    cache_path = db.Column(db.String(34), nullable=True, default=None)
```
yarikoptic, line 29: So what this `cache_path` be now?  Please describe briefly in a comment above, so it also becomes clear why `34` is a sufficient length.

candleindark: `cache_path` will be a relative path of 

```python
        uuid = uuid4().hex
        path = Path(uuid[:3], uuid[3:6], uuid[6:])
```

The path relative to the directory specified by `DATALAD_REGISTRY_DATASET_CACHE`, the base cache path.

datalad_registry/utils/misc.py
```diff
@@ -9,3 +13,29 @@ class StrEnum(str, Enum):
     """
 
     pass
+
+
+def allocate_ds_path() -> Path:
```
candleindark, line 18: @yarikoptic This function generates values to be assigned to `cache_path` of the `URL` model.

datalad_registry/models.py
yarikoptic: Great, thank you for the explanation!

We need an an endpoint which , given a `cache_path`  would return a dataset record. And then also a command line entry point for the same, and ideally it should be smart to operate on full path, thus taking the last three path components -- this way user (us) could quickly figure out what dataset a given path corresponds to.

candleindark: It makes complete sense to me to have such a command line entry for us, the operators, to have access to such information conveniently. However, since `cache_path` is of the internal organization of a running instance of datalad-registry. Don't you think it is a bad idea to expose it through the public `API`?

yarikoptic: I don't see a problem with that, in particular if it is just a UUID (random) and you need to know it to get some non 404 response.  What concerns do you have from exposing it?

candleindark: Not particular concern in mind. I just prefer withholding as much internal information as possible from the public as possible.

As an operator, you will, possible always, have access to the command line. Exposing the mapping of dataset URL records and `cache_path` through an API endpoint provides an additional means of access for convenience.

Let me know your final decision on whether you want the command line support alone or both the command line and API endpoint support. I will file an issue correspondingly.

yarikoptic: target use -- command line. I just didn't want to create another app which would interface to the same DB in parallel to the API app.  But if you think it is easier/ok -- go ahead without API component for now.

yarikoptic (approved):
CLI for getting dataset record for a given cache path I guess could be done later. I have glanced only cursory over the PR -- let's better see it in action!

datalad_registry/models.py
candleindark: A new issue is created for this feature. https://github.com/datalad/datalad-registry/issues/165

Issue #161, opened 2023-04-26
Extract task for metadata extraction to be chained outside at celery level


yarikoptic:
ATM cloning of URL triggers new celery task to do metadata extraction. @candleindark wants to use celery primitive `chain` (see https://docs.celeryq.dev/en/stable/userguide/canvas.html#the-primitives) to chain "clone" and "extract-metadata" tasks.

PR #166, opened 2023-05-08
Initiate dataset-level metadata extractions via Celery task composition primitives

candleindark:
This PR groups dataset-level metadata extractions for different extractors into a group of parallel tasks and "chains" this group of tasks to be dependent on the success of the initial processing of the corresponding dataset URL, which includes the cloning of the dataset. Celery task composition primitives has been utilized to accomplish the goal of this PR.

An additional benefit of utilizing Celery task composition primitives is that logging of errors can now be triggered uniformly as a Celery task.

This PR closes https://github.com/datalad/datalad-registry/issues/161.


yarikoptic:
Looks LGTM , thank you!

PR #168, opened 2023-05-10
Modify Datalad-registry client to adopt to API V2

candleindark:
Modify the Datalad-registry client to adopt API V2 of Datalad-registry.

This PR closes https://github.com/datalad/datalad-registry/issues/164


yarikoptic:
looks LGTM. Wasn't sure if addition/use of yarl is really needed, but oh well -- why not, seems handy ;)  Let's try as deployed.

PR #184, opened 2023-06-01
Feature: Make metadata available through the dataset URLs API endpoint

### LINKED ISSUE #176: Add to /api/v2/dataset-urls options to also include metadata in return
We might want to include specific extractors metadata

- `return_metadata: str`: "", "references" (`{"metalad_core": "http://localhost:5000/api/v2/url-metadata/54"}, "content" (`{"metalad_core": {...}`)

based on the value of `return_metadata` the result would populate `"metadata"` field in the return.  so with `return_metadata="references"` you get records sufficient to display what we display now in web UI. Good point to try jqtables (as on datasets.datalad.org)

candleindark:
This PR closes #176.

This PR adds the query parameter of `return_metadata` to the `GET /api/v2/dataset-urls` endpoint. If this query parameter is not provided, the `metadata` field of each returned dataset URL object will be `null`. If this query parameter is "reference", the `metadata` field of each returned dataset URL object will be a list of objects each presenting a reference link to a piece of metadata of the dataset at the URL. If this query parameter is "content", the `metadata` field of each returned dataset URL object will be a list of objects each presenting a piece of metadata of the dataset at the URL.

Note: I used "reference" instead of "references" as a value of `return_metadata` because I feel that this value is an adjective pertaining to how each piece of metadata should be returned.

### DIFF
--- a/datalad_registry/blueprints/api/dataset_urls/models.py
+++ b/datalad_registry/blueprints/api/dataset_urls/models.py
@@ -104,19 +104,16 @@ class QueryParams(BaseModel):
 
     return_metadata: Optional[MetadataReturnOption] = Field(
         None,
-        # todo: description to be modified
         description="Whether and how to return metadata of the datasets at the URLs. "
         "If this query parameter is not provided, the `metadata` field "
         "of each returned dataset URL object will be `null`. "
         'If this query parameter is "reference", '
-        "the `metadata` field of each returned dataset URL object will be "
-        "an object with metadata extractor names being the fields and "
-        "the links to the corresponding metadata being the values. "
+        "the `metadata` field of each returned dataset URL object will be a list of "
+        "objects each presenting a reference link to a piece of metadata "
+        "of the dataset at the URL. "
         'If this query parameter is "content", the `metadata` field '
-        "of each returned dataset URL object will be an object "
-        "with metadata extractor names being the fields "
-        "and objects representing the corresponding metadata "
-        "being the values.",
+        "of each returned dataset URL object will be a list of objects "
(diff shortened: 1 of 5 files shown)


candleindark:
Changes of this PR has been applied to the running instance of Datalad-registry on Typhon.

datalad_registry/blueprints/api/dataset_urls/models.py
```diff
@@ -88,6 +102,20 @@ class QueryParams(BaseModel):
         "in the query.",
     )
 
+    return_metadata: Optional[MetadataReturnOption] = Field(
+        None,
+        description="Whether and how to return metadata of the datasets at the URLs. "
+        "If this query parameter is not provided, the `metadata` field "
+        "of each returned dataset URL object will be `null`. "
```
yarikoptic, line 109: would it be hard to change it that we simply do not return `metadata` field at all?  or in other words why do you prefer to return `null`/`None` instead?

yarikoptic (approved):
In general LGTM, just left one comment on possibly avoiding to remove `metadata` entirely in case if not requested.

yarikoptic (approved):
In general LGTM, just left one comment on possibly avoiding to remove `metadata` entirely in case if not requested.

datalad_registry/blueprints/api/dataset_urls/models.py
candleindark: Now I see that there is an easy way to do it that I didn't realize before. I may just put in that change if I think it doesn't effect the project negatively in the long run.

yarikoptic:
Let's merge and if to RF for `metadata` key to be absent if not requested -- can be a follow up PR.

PR #190, opened 2023-06-15
Improvement: Skip `null` fields in API responses

candleindark:
This PR modifies the API so that no field having the value of `null` will be returned as part of an API response. A field that has the value of `null` in an API response before this PR will not be contained in the comparable API response after this PR.

This PR resolves the issue brought up in https://github.com/datalad/datalad-registry/pull/184#discussion_r1221783199.

Changes in this PR have been applied to the running instance of datalad-registry on Typhon


yarikoptic:
I had only `metadata` field in mind, but I guess good to be uniform. Thank you!

Issue #200, opened 2023-06-27
add "datetime" into elements of the "branches" record


yarikoptic:
This would allow to see where the most recent work was done (may be just in 'git-annex' branch which is the availability information only)

ATM, looking at the test

```
@pytest.mark.parametrize(
    "ds_name",
    [ 
        "empty_ds_annex",
        "two_files_ds_annex",
        "empty_ds_non_annex",
        "two_files_ds_non_annex",
    ],
)
def test_get_origin_branches(ds_name, request, tmp_path):
    ds: Dataset = request.getfixturevalue(ds_name)
    ds_clone = clone(source=ds.path, path=tmp_path)

    origin_branches = get_origin_branches(ds_clone)
    
    assert origin_branches == [
        {"name": branch_name, "hexsha": ds.repo.get_hexsha(branch_name)}
        for branch_name in ds.repo.get_branches()
```

we have only "name" and "hexsha" and greedy me wants to add also the datetime for that "hexsha" commit.

candleindark:
@yarikoptic I completed and merged #201. I think this issue can be closed now, but I can't be sure completely. Let me know if I have missed anything else to be done for this issue. Otherwise, let's close this issue.

PR #201, opened 2023-06-27
Add authordate to branch record

yarikoptic:
On the way to address #200


yarikoptic:
cool. eventually we might need to do a query and request "forceful refresh" so we re-extract those dates.

Issue #221, opened 2023-07-25
Bug: Remove the bug that is inherent in the current Celery app setup


candleindark:
The current setup of the Celery app in the project produces different Celery app instances with inconsistent settings/configurations depending on the order the modules are imported. In particular, the task functions defined in `tasks.py` can be executed in the Flask app's context or not depending on the order those tasks function are imported in relation to creation of the Flask app. This is a problem because a task function may end up running without the Flask app's context when the context is needed, and the context may be ended up loaded more than once, resulting in corruptions.

This issue can be observed in some of the tests of the Celery tasks where the context of the Flask app has to be brought up explicitly to use a celery task. For example: https://github.com/datalad/datalad-registry/blob/8cb28e244f681895343321ea6b7ce0a930de04f5/datalad_registry/tests/test_tasks/test_extract_ds_meta.py#L15-L25

On the other hand, Celery tasks can bring up the Flask app's context by themselves in other usages. For example, https://github.com/datalad/datalad-registry/blob/8cb28e244f681895343321ea6b7ce0a930de04f5/datalad_registry/blueprints/api/dataset_urls/__init__.py#L49-L117

This issue can be resolved by integrating the Celery app with the Flask app in the project in the manner suggested by by [Flask's documentation](https://flask.palletsprojects.com/en/2.3.x/patterns/celery/).

candleindark:
The entire Celery app setup process has been re-written as part of #222, so this bug is fixed.

Issue #195, opened 2023-06-19
Provide a "read-only" public instance


yarikoptic:
For starters it should probably have only the "query" interfaces made available, so it would not be able to trigger any action/update database.

Think about how could this be done?

TODOs
- [ ] @yarikoptic -- do provide test "public URL" which would scope only to overview and `api/v2/url-metadata`
- [x] @candleindark 
  - [x] make it switchable to enable POST request to report a URL to consider
  - [x] run non-devel instance
  - [x] make celery turnable off 
  - [x] replication of DB so we could do stuff on one instance and just expose information read-only to the world

yarikoptic:
this is done AFAIK -- we have http://registry.datalad.org/

Issue #112, opened 2023-03-01
"cron job" to keep dataset updated


yarikoptic:
I sorted by "Last update" to get a few which seems never got analyzed but the sample one I went to exists and well

URL | Dataset | Commit | Annex keys | Annexed files in working tree | Size of Annexed files in working tree (bytes)
-- | -- | -- | -- | -- | --
https://github.com/anthfm/afm.git | None |   |   | None | None |   |  
https://github.com/ReproNim/ReproTube.git | None |   |   | None | None |   |  
https://github.com/dguibert/nur-packages.git | None |   |   | None | None |   |  
https://github.com/adswa/ml-books-gitannex.git | None |   |   | None | None |   |  
https://github.com/datalad-datasets/machinelearning-books.git | None |   |   | None | None |   |  
https://github.com/anthfm/s3-test.git | None |   |   | None | None |   |  
https://github.com/anthfm/dd2.git | None |   |   | None | None |   |  
https://github.com/soichih/datalad-o3d-test.git | None |   |   | None | None |   |  
https://github.com/TrueTube/Andriy_Popyk.git | None

I even wonder if we should add some "Last update error" DB field to store information if it was some errored out execution and add it as the last column to web ui table to ease troubleshooting... WDYT?

candleindark:
I am aware of this. Last time I checked, the logic in handling update was not fully implemented. The idea of using a central dispatcher with an dedicated queue I mentioned today should solve the problem, but I may use a more expedient solution for now.

yarikoptic:
This is suggested changes to the model which should IMHO facilitate all desired operations for checks more or less efficiently

```patch
❯ git diff
diff --git a/datalad_registry/models.py b/datalad_registry/models.py
index 75c0faf..b9748d9 100644
--- a/datalad_registry/models.py
+++ b/datalad_registry/models.py
@@ -15,13 +15,20 @@ class URL(db.Model):  # type: ignore
     annex_key_count = db.Column(db.Integer)
     annexed_files_in_wt_count = db.Column(db.Integer)
     annexed_files_in_wt_size = db.Column(db.BigInteger)
-    info_ts = db.Column(db.DateTime(timezone=True))
-    update_announced = db.Column(db.Boolean, default=False, nullable=False)
     head = db.Column(db.Text)
     head_describe = db.Column(db.Text)
+    head_dt = db.Column(db.DateTime(timezone=True))
     branches = db.Column(db.Text)
     tags = db.Column(db.Text)
[...]

PR #222, opened 2023-07-31
Overhaul setup of Flask and Celery apps

candleindark:
This PR overhauls the setup of the Flask and Celery apps (including the integration of Celery app into the Flask app).

This PR is created to address numerous issues, in particular:

1. #221
2. Incorporate the solution in #96 to make a result backend available in the Celery app
3. Provide a solid foundation to cleanly set up Celery beat service for #112 
4. Provide a high level configuration switch with pre-loaded defaults and requirements for different modes of operation, `PRODUCTION`, `DEVELOPMENT`, `TESTING`, `READ_ONLY`. In particular, the implementation of the `READ_ONLY` mode will be the solution for #195.

It also has some additional benefits such as:
1. Validating and parsing configurations, provided through env vars, in the correct type
2. Ensuring the app only starts after all the needed configurations are provided through env vars


.env
```diff
@@ -1,2 +0,0 @@
-export DATALAD_REGISTRY_INSTANCE_PATH=$PWD/instance
-export DATALAD_REGISTRY_DB_INSTANCE_PATH=$PWD/instance/db
```
yarikoptic, line 2: could you please complement this removal with providing a file (or files) instead of this which would setup system in one of those configurations.

candleindark: `template.env.dev` and `template.env.testing` are now provided. They are to be used with `docker-compose.dev.yml` and `docker-compose.testing.yml` respectively.

template.env.dev
candleindark, line 1: I think we should revert this one.

I could have done something similar in constructing `CELERY_BROKER_URL`, `SQLALCHEMY_DATABASE_URI`, and `FLOWER_BROKER_API` in the docker compose file. I didn't because doing simple constructions this way doesn't allow proper escaping of special characters in usernames and passwords. I even left a comment for the user as "Make sure user name and password characters are escaped properly for URL format".

candleindark (commented):
how can I delete a comment here?

template.env.dev
yarikoptic: well, you can also just improve the comments that passwords and logins should be chosen such that they do not require URL escaping or if a tricky one is needed - people would need to replace them with urlencoded in the URLs. 
But here duplication is IMHO more evil than adding those "restrictions" since while duplicating you are adding more chances for misery.  And I genuinely have allergy to such duplications.

candleindark: Yes, you are right. Modifying the comment is a much better way to go.

candleindark (commented):
@yarikoptic After [digger deeper](https://docs.docker.com/compose/compose-file/13-merge/), I was able to find a way to use just one overriding Docker Compose file to mount the local current directory at the host to the web service container for development. So, I am deleting the two overriding files that were proposed initially. I hope this is OK for you.

yarikoptic:
sounds good but please then immediately adjust documentation we changed to reflect that: https://github.com/datalad/datalad-registry/pull/222/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R44

candleindark:
> sounds good but please then immediately adjust documentation we changed to reflect that: https://github.com/datalad/datalad-registry/pull/222/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R44

Done

yarikoptic:
let's then just merge it and proceed from there!

PR #223, opened 2023-08-16
Simplify flask app fixture

candleindark:
After recent changes involving the rewriting of the setup Flask and Celery apps and docker environment, #222 for example, a simple fixture with function scope for setting up the Flask app works fine now. The former problem of connections to the database not being relinquished, possibly by completed tests in a test session, no longer exists.

This simplified fixture is actually better for ensuring the Flask and Celery apps are setup correctly with a new storage state (the state of cache, instance path, database, etc.)

Note:
The branch for this PR should be merged with master after #222 is merged into master to have its relevant commit reflected in this PR.


candleindark:
While this solution is able relinquish the connections to Postgres DB it seems to hold on the connections to Redis, the result backend. 

Many attempts have been made to close the connection to Redis manually without success, and the documentation from Celery regarding connections is pretty poor. Additionally, this can the result of an existing bug in Celery, https://github.com/celery/celery/issues/4465.

I am giving up this PR and continue to the use the Flask and Celery apps fixture setup that shares the same Flask and Celery apps for the entire test session for the time being.

Issue #134, opened 2023-03-14
Why last-update 	 2023-02-23T02:00:34+0000 ?  Adjust update strategy


yarikoptic:
I highly doubt that there were no changes to any of the repositories since then.  Don't we run some cron job which tests if there are updates to already known datasets? obviously we might be not running enough of updates but we should get some and we should do them by looking first at the least recently updated -- i.e. with the oldest `info_ts`, and thus result in them being updated even if there is no actual updates from the remote.

candleindark:
There is a cron job running, but I don't think the update logic is right as mentioned in https://github.com/datalad/datalad-registry/issues/136. I will put in some changes to make update working starting with re-organizing the local cache as discussed in https://github.com/datalad/datalad-registry/issues/125.

candleindark:
PR #238 closes this issue as well.

yarikoptic:
if it does -- add `Closes #238` in the description of that PR and the issue would be closed when that PR is merged

PR #238, opened 2023-09-05
Update cron job

candleindark:
This PR is intended to provide the implementation of a periodic updater that fulfills that intended goal for a periodic updater described in issue #112, and it closes both #112 and #134.

In support of the main goal, this PR also modifies existing Celery tasks, `process_dataset_url` and `extract_ds_meta`, so that they appropriately lock the `RepoUrl` record they modify to avoid potential race conditions and handle situations in which the given dataset URL id argument no longer identifies a `RepoUrl` record in the database because of deletion before the execution of the tasks.

TODOs
- [x] add some tests to ensure at least identifying which repos are to be updated; unittest on actual update function

### DIFF
--- a/datalad_registry/tasks/__init__.py
+++ b/datalad_registry/tasks/__init__.py
@@ -170,7 +170,7 @@ def extract_ds_meta(ds_url_id: StrictInt, extractor: StrictStr) -> ExtractMetaSt
     validate_url_is_processed(url)
 
     # Absolute path of the dataset clone in cache
-    cache_path_abs = url.cache_path_absolute
+    cache_path_abs = url.cache_path_abs
     assert cache_path_abs is not None
 
     # Check for missing of required files
@@ -296,7 +296,7 @@ def process_dataset_url(dataset_url_id: StrictInt) -> ProcessUrlStatus:
         # === there is no RepoUrl in the database with the specified ID ===
         return ProcessUrlStatus.NO_RECORD
 
-    old_cache_path_absolute = dataset_url.cache_path_absolute
+    old_cache_path_absolute = dataset_url.cache_path_abs
 
     # Allocate a new path in the local cache for cloning the dataset
     # at the specified URL
--- a/datalad_registry/tasks/__init__.py
+++ b/datalad_registry/tasks/__init__.py
@@ -459,6 +459,8 @@ def url_chk_dispatcher() -> list[int]:
         .limit(max_chks_to_dispatch)
     ).all()
 
+    db.session.rollback()  # Release the lock
+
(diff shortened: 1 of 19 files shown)


datalad_registry/tasks.py
```diff
@@ -299,3 +301,157 @@ def process_dataset_url(dataset_url_id: StrictInt) -> None:
...
+                    RepoUrl.chk_req_dt.is_not(None),
+                    RepoUrl.n_failed_chks <= max_failed_chks,
+                    or_(
+                        # The ones that have not been checked since the request
+                        RepoUrl.last_chk_dt.is_(None),
+                        RepoUrl.last_chk_dt < RepoUrl.chk_req_dt,
+                        # The ones that have been checked since the request but
+                        # the last check is old enough
+                        RepoUrl.last_chk_dt <= age_cutoff,
+                    ),
```
candleindark, line 340: @yarikoptic I simplified the logic for selection here.

yarikoptic:
note -- tests are still failing.

candleindark:
> note -- tests are still failing.

It was caused by some mypy complain. I had some pseudocode in `chk_url` removed, and the tests are passing now.

datalad_registry/tasks/__init__.py
```diff
@@ -299,3 +344,199 @@ def process_dataset_url(dataset_url_id: StrictInt) -> None:
...
+
+    # ===
+    # The RepoUrl has not been checked by any other process
+    # since this check was initiated
+    # ===
+
+    # todo: the following is just a template
+    try:
+        # Do the checking and potential updating
+        pass
```
yarikoptic, line 528:
```
# if DB information is different from what we have
if url.head != ds.repo.get_hexsha():
    # trigger the metadata and stats re-extraction
```

and may be remove that `"origin/HEAD"` where you now do `get_hexsha` and `.describe`

datalad_registry/tasks/__init__.py
```diff
@@ -0,0 +1,556 @@
...
+    """
+
+    # Get the RepoUrl from the database by ID with a read/share lock
+    url = (
+        db.session.execute(
+            select(RepoUrl).filter_by(id=ds_url_id).with_for_update(read=True)
+        )
+        .scalars()
+        .one_or_none()
+    )
```
yarikoptic, line 153: smells like having some helper which absorbs this pattern might be worthwhile (not in this PR, generally)

```python
url = get_url(ds_url_id, read=True, none_ok=True)
```

candleindark:
FYI: A running instance of this version of Datalad-registry is available on Typhon

PR #246, opened 2023-09-26
Add basic DANDI metadata "extraction"

yarikoptic:
Not even really extraction since we will just load stored in DataLad dandisets .yaml (and later .json files for assets).

TODO:

- [ ] replace prototyped code with proper python (asset level metadata -- postpone for now)
- [ ] deploy and see what happens -- in principle we should also be able to search through dandisets


candleindark:
This PR is replaced by #257.

PR #252, opened 2023-10-09
Read-only instance

candleindark:
This PR is intended to fulfill #195.

This PR is based on #238. To get the relevant commit for this PR, merge the master branch into the associated branch after #238 is merged into the master branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,104 @@ The `instance` directory for the web service is in the current directory at the 
 
 *Note*: Other services will not react to the changes in the codebase at the host machine.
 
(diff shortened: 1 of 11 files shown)


candleindark:
@yarikoptic A read-only instance of Datalad-Registry is now on Falkor.

candleindark:
Per @yarikoptic's recommendation, it may be a good idea to respond `405 Method Not Allowed` instead of `418 I'm a teapot` in the event of a user posting to the `declare_dataset_url` endpoint.

PR #256, opened 2023-10-11
Restrict Flask to version below 3.0

candleindark:
The latest version of Flask, 3.0, doesn't work with the existing code. Let's restrict the version of Flask to be below 3.0 for now and update the code base to support Flask 3.0+ in another occasion.


candleindark:
The change in this PR has been included in #255 instead (so that the tests can be passed).

Issue #129, opened 2023-03-13
extract/store metadata for dandisets


yarikoptic:
- [ ] needs metadata extractor to be provided . This datalad extension already depends on metalad, I don't think there is sense in developing dedicated dandi datalad extension ATM
   - all DataLad dandisets have metadata provided in  (example on randomly chosen 000402) [dandiset.yaml](https://github.com/dandisets/000402/blob/draft/dandiset.yaml) for dataset level, and then [.dandi/assets.json](https://github.com/dandisets/000402/blob/draft/.dandi/assets.json) for each asset.  So for starters it is just a matter of loading them all up
- [ ] identify dandisets by presence of `dandiset.yaml` in the top directory

candleindark:
I have the following questions/statements to be confirmed/answered.

1. A Dandiset is a Datalad dataset. [True/False]
2. Assuming the last statement is True. A proper Dandiset contains a `dandiset.yaml` directly under. [True/False]
3. Does every Dandiset contain a `.dandi/assets.json`?
4. Assuming 1 and 2 are both true. For the extraction of metadata from a Dandiset, you want to load the data from the `dandiset.yaml` and `.dandi/assets.json` (if `.dandi/assets.json` exists`) and save the loaded data to the database in JSON format. [True/False]
5. Assuming 4 is true. How do you want to combine the loaded data from both `dandiset.yaml` and `.dandi/assets.json`? Do you want to put the two pieces of data under different keys? If yes, what keys?
6. You want datalad-registry to extract the aforementioned metadata from a Dandiset directly without going through the `dataset metalad` extension, which does not support such extraction at this point. [True/False]

candleindark:
@yarikoptic Is it possible for the "dandiset.yaml" file, as a document stream, to contain more than one document?

yarikoptic:
that should not happen. It would be ok to error out if it does

candleindark:
This issue has been closed by #257.

PR #257, opened 2023-10-13
Basic DANDI metadata extraction

candleindark:
This PR provides a solution for extracting basic metadata from DANDI datasets. It fulfills the requirements specified in #129 and replaces #246 (#246 is unfortunately too many commits behind master branch).

The solution in the PR is based on the outlined solution provided in #246.

This PR is based on #254. Thus, to get the relevant commits of this PR, merge master branch to the associated branch of this PR after #254 is merged to the master branch.


candleindark:
Changes from this PR have been applied to the running instance of Datalad-registry on Typhon.

datalad_registry/conf.py
```diff
@@ -36,6 +36,9 @@ class BaseConfig(OperationConfig):
         "metalad_studyminimeta",
         "datacite_gin",
         "bids_dataset",
+        # === DANDI related extractors ===
+        "dandi:dandiset",
+        # "dandi:assets",  # Let's not activate this yet by default
```
yarikoptic, line 41: Let's go (here or right after) for agreeing on "semantic" that without `:SMTH` it is dataset level metadata, and then with `:files` it would be the list with per-file metadata.

```suggestion
        "dandi",
        # "dandi:files",  # Let's not activate this yet by default
```

`files` since it is per "file" metadata (although .zarr would be a folder) and "asset" is an internal DANDI term, as well as "dandiset".

yarikoptic (approved):

PR #260, opened 2023-10-19
Clear instance path in flask app fixture

candleindark:
This PR ensures that the flask app fixture used in tests provides a clean instance folder for the Flask app initially.

### DIFF
--- a/datalad_registry/tests/conftest.py
+++ b/datalad_registry/tests/conftest.py
@@ -55,10 +55,12 @@ def flask_app(_flask_app):
         db.drop_all()
         db.create_all()
 
-    # Reset the base local cache for datasets
+    # Clear the instance folder of the Flask app and the base local cache for datasets
+    instance_path = Path(_flask_app.instance_path)
     cache_path = _flask_app.config["DATALAD_REGISTRY_DATASET_CACHE"]
-    rm_ds_tree(cache_path)
-    cache_path.mkdir()
+    for p in [instance_path, cache_path]:
+        rm_ds_tree(p)
+        p.mkdir()
 
     celery_app: Celery = _flask_app.extensions["celery"]
 


datalad_registry/tests/conftest.py
```diff
@@ -55,10 +55,12 @@ def flask_app(_flask_app):
...
 
-    # Reset the base local cache for datasets
+    # Clear the instance folder of the Flask app and the base local cache for datasets
+    instance_path = Path(_flask_app.instance_path)
     cache_path = _flask_app.config["DATALAD_REGISTRY_DATASET_CACHE"]
-    rm_ds_tree(cache_path)
-    cache_path.mkdir()
+    for p in [instance_path, cache_path]:
+        rm_ds_tree(p)
+        p.mkdir()
```
yarikoptic, line 63: interesting... why during tests those paths are not under some temporary folder?

candleindark: > interesting... why during tests those paths are not under some temporary folder?

Those folders are in a temporary directory except that the temporary directory is session-scoped. 
https://github.com/datalad/datalad-registry/blob/c9a9f5aa5ac30a9c30bca80cd199dd23cbdf6b56/datalad_registry/tests/conftest.py#L18-L72. Essentially, I reuse the same Flask app across the tests. I attempted to use a new Flask app for each test, but we experience exhausting the connections to the database. (I actually, make numerous attempt to solve the problem without success). 

This aspect of reusing the same Flask app across the tests is actually the same as before I get started with this project. I still think having a new Flask app for each test is a better way to go, but for the time being, the current setup serve us well enough. (It is actually faster as well.)

yarikoptic: well, it makes me then realize that we then cannot run tests in parallel since fixture would keep kicking out cache from under the test's feet... is there some explicit way - a somewhat related https://github.com/pytest-dev/pytest-xdist/issues/84 . Let's proceed for now

yarikoptic (approved):

datalad_registry/tests/conftest.py
candleindark: I think this aspect of the code wouldn't prevent us from running the tests in parallel.

According to ChatGPT,

"In pytest, a session-scoped fixture is intended to be created only once per test session. However, when using `pytest-xdist` to run tests in parallel, the concept of a "session" becomes a bit more complex.

### Session-Scoped Fixtures in xdist

With `pytest-xdist`, each worker (subprocess) spawned for parallel execution gets its own "session," and therefore, its own session-scoped fixtures. This means that session-scoped fixtures do not get shared across multiple workers in `pytest-xdist`.

This is critical to understand because it might have implications for resource allocation, like database connections, temporary files, etc. For example, if you have a session-scoped fixture that sets up a database, each worker will set up its own separate database."

Because each worker is getting its own session-scoped fixtures, each worker will get its own Flask app, Flask instance directory, and dataset cache directory. Emptying out each instance and cache directories is just the respective worker's business.

PR #267, opened 2023-10-27
Specify the version of the annexremote dependency in requirements.txt

candleindark:
This allows finer control over which version of annexremote is installed in the Python environment.


candleindark:
The problem with the new release of annexremote is now fixed. Let forget about specifying specific version of annexremote in requirements.txt for now.

PR #275, opened 2023-11-24
Eliminate shellcheck err in by adding `-r` option

candleindark:
This PR eliminate the shellcheck complain about calling `read` without the `-r` option.

We should be able to call `read` with the `-r` option in the use case since URLs can't contain `"\"`.


yarikoptic:
Thank you @candleindark - I fell into my own CI trap by committing without a PR ;)

PR #276, opened 2023-11-24
Worker pids limit

candleindark:
This PR provides a workaround solution to set the PID limit of the worker service container. This workaround is needed because currently Podman Compose, as of version `1.0.6`, doesn't support setting of the PID limit by the [`pid_limit` key](https://docs.docker.com/compose/compose-file/05-services/#pids_limit) or the [`deploy.reservations.pids` key](https://docs.docker.com/compose/compose-file/05-services/#pids_limit).

This solution provides `docker-compose.dev.no-worker.yml` to bring up all services in Datalad-Registry except the worker service and `worker-up.sh` to bring up the worker service with option to choose the PID limit using the `--pids-limit` option.

Note: 
1. A request has been made to the `podman-compose` project to add support of setting the PID limit of a service container. (https://github.com/containers/podman-compose/issues/806)
2. Additionally, this workaround will not be needed as soon as `Podman` is update to version 4.4 or later, https://docs.podman.io/en/v4.4/markdown/podman-update.1.html#pids-limit-limit.

### DIFF
--- a/docker-compose.dev.no-worker.yml
+++ b/docker-compose.dev.no-worker.yml
@@ -1,0 +1,130 @@
+services:
+  web:
+    build:
+      dockerfile: dev.Dockerfile
+      ## Old docker-compose asked for context but this was still insufficient
+      ## since it could then "link" the built image here to be used for the worker
+      # context: .
+    image: datalad-registry:dev
+    depends_on:
+      broker:
+        condition: service_healthy
+      db:
+        condition: service_healthy
+    ports:
+      - "5000:5000"
+    environment: &env
+      FLASK_APP: "datalad_registry:create_app"
+
+      DATALAD_REGISTRY_OPERATION_MODE: "${DATALAD_REGISTRY_OPERATION_MODE}"
+      DATALAD_REGISTRY_INSTANCE_PATH: /app/instance
+      DATALAD_REGISTRY_DATASET_CACHE: /data/cache
+
+      CELERY_BROKER_URL: "${CELERY_BROKER_URL}"
+      CELERY_RESULT_BACKEND: "redis://backend:6379"
+
+      # db service access info
+      SQLALCHEMY_DATABASE_URI: "${SQLALCHEMY_DATABASE_URI}"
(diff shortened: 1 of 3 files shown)


yarikoptic:
Thank you for tracking it all down.  Good that you shared the `ps` output which finally let us see the actual reason for the need for so many PIDs - they were not picked up. We had similar issue awhile back in buildbot setup where we also had clients start many of processes in docker container. Pasting here some references from my reply on matrix

Here is the commit where I added a simple init process to fix it  https://github.com/datalad/buildbot/commit/3dc7cb17e05768ba1decbe3161a2405ab2ef986e
Here more on that zombie ripper https://github.com/phusion/baseimage-docker/tree/master#whats-inside-the-image

So I think we should not bother raising PID limit, but rather add this `init` process into the setup

yarikoptic:
ok, if easy to rebase image to be on top of that https://github.com/phusion/baseimage-docker/tree/master -- go ahead.

Otherwise -- we could just copy that https://github.com/phusion/baseimage-docker/blob/master/image/bin/my_init here, and use it instead of plain "bash -c" to start celery worker. Something like 

```patch
diff --git a/docker-compose.dev.yml b/docker-compose.dev.yml
index 97a9974..f40bc43 100644
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -37,10 +37,11 @@ services:
       db:
         condition: service_healthy
     command: [
-      "bash", "-c",
-      "celery -A datalad_registry.make_celery:celery_app worker --loglevel INFO --pool prefork"
+      "my_init", "--skip-startup-files", "--skip-runit", 
+      "celery", "-A", "datalad_registry.make_celery:celery_app", "worker", "--loglevel", "INFO", "--pool", "prefork"
     ]
     volumes:
+      - ./bin/my_init:/usr/local/sbin/my_init
       - ${WORKER_PATH_AT_HOST}/data/cache:/data/cache
     environment:
       <<: *env
```

candleindark:
This approach has been abandon. We will ensure the Docker image for the project has a correct init process instead which is accomplished by #282.

PR #279, opened 2023-12-17
make ds id column value into a hyperlink to search for that uuid

yarikoptic:
Makes it easy to navigate the "forks" of a given dataset -- pretty neat.

1st commit is needed to get it working locally, a version of it is also in #278


candleindark:
LGTM.

PR #282, opened 2023-12-22
Base Docker image of project on phusion/baseimage

candleindark:
Base Docker image of project on phusion/baseimage. It has the correct init process that ensures all orphaned child processes are reaped.


yarikoptic:
Is that enough? I thought you would need to add script to start the desired process under that init process

candleindark:
You are right. That's not enough. I will put out a correction later.

Issue #283, opened 2024-01-02
Ensure all all orphaned child processes are reaped in the worker service


candleindark:
As pointed out in https://github.com/datalad/datalad-registry/pull/282#issuecomment-1867307788, basing the Docker image of the project on phusion/baseimage simply can't ensure all all orphaned child processes are reaped in the worker service. Adjust `dev.Dockerfile` so that all all orphaned child processes are reaped in the worker service.

Issue #271, opened 2023-11-17
Automate (cron job script) getting updates from the datalad-usage-dashboard


yarikoptic:
ATM (@candleindark correct me if I am wrong) the datasets we have known to registry came from a single "injection" of them into DB using https://github.com/datalad/datalad-registry/blob/master/tools/populate.py . 
That script seems to not care about "updates only mode".  So ideally, to not stress registry more than needed, we should not "post" on old URLs which it already knows.  

Thus likely we should add mode there to first

- get a list of already known URLs from registry
- POST only for new ones

and place that script on cron to run daily on the main server.

candleindark:
@yarikoptic You are right. That script was initially written with the purpose to populate an empty Datalad-Registry.

Instead of placing that script on cron, can I implement the updates from datalad-usage-dashboard in our Celery scheduler within Datalad-Registry instead?

yarikoptic:
Sure! I forgot that we have a formal schedule within the system

PR #284, opened 2024-01-03
Ensuring container services having the correct init process and other improvements related to container services

candleindark:
This PR contains the following improvements in container services specification.

1. Container services are now utilizing the init process provided by https://github.com/phusion/baseimage-docker to ensure all orphaned processes are reaped. Thus, this PR closes https://github.com/datalad/datalad-registry/issues/283.
2. Elimination of the calls to `bash` in invoking the command to be executed for a container when the calls can be avoided.
3. The use of `exec` to ensure a `SIGTERM` signal is received by the target process of a container when `docker/podman stop` are used.
4. Upgrading the operating system inside the container when building the container images. This improve the security of the container. See https://github.com/phusion/baseimage-docker?tab=readme-ov-file#upgrading-the-operating-system-inside-the-container for more information

### DIFF
--- a/docker-compose.dev.local.override.yml
+++ b/docker-compose.dev.local.override.yml
@@ -4,8 +4,11 @@
 
 services:
   web:
-    command: [ "bash", "-c",
-               "git config --global --add safe.directory /app && pip3 install -U -e . && flask init-db && flask run --host=0.0.0.0 --debug" ]
+    command: [
+      "/sbin/my_init", "--",
+      "bash", "-c",
+      "git config --global --add safe.directory /app && pip3 install -U -e . && flask init-db && exec flask run --host=0.0.0.0 --debug"
+    ]
     volumes:
       - ./:/app
       - ./instance:/app/instance
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -9,6 +9,7 @@ WORKDIR /app
 # TODO: Consider removing the eatmydata dependency. It may not be needed.
 RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends eatmydata && \
+    DEBIAN_FRONTEND=noninteractive apt-get upgrade -y -o Dpkg::Options::="--force-confold" && \
     DEBIAN_FRONTEND=noninteractive eatmydata apt-get install -y --no-install-recommends gnupg locales && \
     echo "en_US.UTF-8 UTF-8" >>/etc/locale.gen && locale-gen && \
     DEBIAN_FRONTEND=noninteractive eatmydata apt-get install -y --no-install-recommends \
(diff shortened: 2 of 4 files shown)


docker-compose.dev.local.override.yml
```diff
@@ -4,8 +4,11 @@
 
 services:
   web:
-    command: [ "bash", "-c",
-               "git config --global --add safe.directory /app && pip3 install -U -e . && flask init-db && flask run --host=0.0.0.0 --debug" ]
+    command: [
+      "/sbin/my_init", "--",
```
yarikoptic, line 8: why just not to retain the same ENTRYPOINT of `docker.io/phusion/baseimage:jammy-1.0.1` which is the `/sbin/my_init` AFAIK and then just have here the `"bash", "-c", ...`?

yarikoptic: but otherwise -- looks good if that works ;)  please make it sweat (run the population script again) and see if pids are staying under control with no zombies

candleindark: > why just not to retain the same ENTRYPOINT of `docker.io/phusion/baseimage:jammy-1.0.1` which is the `/sbin/my_init` AFAIK and then just have here the `"bash", "-c", ...`?

The `docker.io/phusion/baseimage:jammy-1.0.1`  image actually doesn't have an ENTRYPOINT of `"/sbin/my_init"`. It only have `CMD ["/sbin/my_init"]` as the default command to execute, which is to be overridden.

I have considered adding `ENTRYPOINT ["/sbin/my_init"]` to our project image, but that would require starting each command to be executed with `"--"`. For example, the command you referenced above would have to be `command: [
      "--",
      "bash", "-c",
      "git config --global --add safe.directory /app && pip3 install -U -e . && flask init-db && exec flask run --host=0.0.0.0 --debug"
    ]`. I think that makes the code less clear for it is not immediately clear what the `"--"` is for, and one has to remember to use `"--"` before specifying any command to be executed because of a hidden context (the entry point) in the project Docker image.

On the other hand, we can add ``ENTRYPOINT ["/sbin/my_init", "--"]` to our project image and specify commands to be executed the way you suggested. However, doing so makes passing options to `"/sbin/my_init"` much less convenient (one has to override the entry point to pass any options to `"/sbin/my_init"` itself).

Because of the above considerations, I think we should keep the current arrangement. It is a bit longer, but it is explicit and convenient to pass options to `"/sbin/my_init"` itself.

candleindark: > please make it sweat (run the population script again) and see if pids are staying under control with no zombies

I want to sweat the server by deploying a solution for https://github.com/datalad/datalad-registry/issues/271 instead. I think we can see zombies if this PR doesn't work when adding all the gin datasets to the registry (the worker process has already accumulated some zombies over the vacation, about 40, just by updating the registered datasets).

yarikoptic: ok

candleindark:
Attached is a snapshot of output of `podman stats` (note the PIDS column) and a snapshot of the `top` output inside the worker container after the changes in the PR is put is put in place.

Screenshot 2024-01-04 at 4 11 18 PM

Screenshot 2024-01-04 at 4 18 24 PM

These images can be used to as a reference for comparison in the future.

PR #287, opened 2024-01-12
Sync Datalad-Registry with datalad-usage-dashboard

### LINKED ISSUE #271: Automate (cron job script) getting updates from the datalad-usage-dashboard

candleindark:
This PR does the following.

1. Implements a periodic Celery task for ensuring all active repos presented in [datalad-usage-dashboard](https://github.com/datalad/datalad-usage-dashboard) are registered in Datalad-Registry (excepts repos from OSF which will be supported later). This PR closes #271.
2. Provides a script to invoke the above Celery task on demand. Thus, an empty instance of Datalad-Registry can be populated with all the active repos presented in the usage dashboard by running this script. 
3. Refactor some of the Pydantic models used by `tools.populate.py` so that these models can be used by the aforementioned Celery task as well.
4. [`responses`](https://github.com/getsentry/responses) will be added as a dependency for testing for the purpose of mocking out the `requests` library.

Note: Changes of this PR has been applied to the Typhon instance.

### DIFF
--- a/datalad_registry/tasks/__init__.py
+++ b/datalad_registry/tasks/__init__.py
@@ -640,7 +640,7 @@ def chk_url_to_update(
 
 class FailedSubmission(TypedDict):
     """
-    A TypedDict a failed submission of a repo URL to Datalad-Registry
+    A failed submission of a repo URL to Datalad-Registry
     """
 
     status_code: int
--- a/datalad_registry/tasks/__init__.py
+++ b/datalad_registry/tasks/__init__.py
@@ -18,7 +18,6 @@ import requests
 from sqlalchemy import and_, case, not_, or_, select
 from yarl import URL
 
-from datalad_registry.blueprints.api.dataset_urls import DATASET_URLS_PATH
 from datalad_registry.com_models import MetaExtractResult
 from datalad_registry.models import RepoUrl, URLMetadata, db
 from datalad_registry.utils import StrEnum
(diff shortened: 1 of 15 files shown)


datalad_registry/tasks/__init__.py
```diff
@@ -631,3 +636,101 @@ def chk_url_to_update(
         db.session.commit()
 
     return ChkUrlStatus.OK_UPDATED if is_record_updated else ChkUrlStatus.OK_CHK_ONLY
+
+
+class FailedSubmission(TypedDict):
```
yarikoptic, line 641: why `TypedDict` and not `dataclass`es which seems to be more common construct pattern to use?

datalad_registry/tasks/__init__.py
```diff
@@ -631,3 +636,101 @@ def chk_url_to_update(
...
+    """
+    A TypedDict a failed submission of a repo URL to Datalad-Registry
```
yarikoptic, line 643: not a sentence... not sure if worth stating TypedDict type here instead of purpose only

yarikoptic:
@jwodder please review

datalad_registry/tasks/__init__.py
candleindark: The reason I used type `TypedDict` is because a dictionary displays better in the Flower interface.

datalad_registry/tasks/__init__.py
candleindark: Yeah, stating `TypedDict` here is unnecessary. The docstring has been updated.

jwodder (approved):

candleindark:
@yarikoptic Is it OK to merge now?

yarikoptic:
sure... I didn't fully grasp why we need so many changes but if it accomplishes the drill -- let's proceed!

candleindark:
> sure... I didn't fully grasp why we need so many changes but if it accomplishes the drill -- let's proceed!

I think the additional changes you are referring to are those resulted from the refactoring of the `tools/populate.py`, testing with `responses` for mocking the `requests` calls, and the additional script to invoke the Celery tasks on demand.

PR #293, opened 2024-01-22
Update the codecov badge link

candleindark:
The original link shows an "unknown" status.


candleindark:
This seems to be a problem with CodeCov's service. The existing badge works fine now after CodeCov reran, so this PR is no longer needed.

PR #296, opened 2024-02-01
Styling facelift

### LINKED ISSUE #290: Tune up main.css to look more like style in catalog and overall not that plain
we have https://github.com/datalad/datalad-registry/blob/2d5e9e296b2edd71b99d36fee997ff64a131eb82/datalad_registry/static/main.css 

as we are hoping to (ab)use catalog, may be we should adopt its styling. But it seems to be more elaborate and uses bootstrap etc already.

@jsheunis WDYT -- is there any hope for an easy facelift of  http://registry.datalad.org/overview/ ? ;-)

jsheunis:
This PR:
- adds css to make the styling similar to datalad-catalog
- adds a datalad-registry logo
- closes #290 

In the example screenshot in the issue, I also change some of the table headings to make the spacing narrower. I did not include these changes in this PR since I was not sure whether that was desired or not, although it can easily be added in another commit.


candleindark:
This is a great improvement already! Thanks @jsheunis. @yarikoptic Do you want @jsheunis to include the changes in the table headings?

yarikoptic:
The narrower the better at this point!  Thank you @jsheunis -- looks great.

jsheunis:
Done

candleindark:
Awesome. @yarikoptic I think this PR is ready to merge. You have the final call. Thanks, @jsheunis .

yarikoptic:
@candleindark would be nice to see that deployed on typhon and readonly falkor

PR #298, opened 2024-02-06
Do provide secret token for CODECOV

yarikoptic:
It is needed for v4 -- we added that secret at the level of the organization so should be available

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,3 +57,6 @@ jobs:
       run: mypy datalad_registry datalad_registry_client
     - name: Upload coverage to codecov
       uses: codecov/codecov-action@v4
+      with:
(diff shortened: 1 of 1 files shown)


yarikoptic:
now I could expect codecov to report coverage back

yarikoptic:
hm, still no codecov reported back -- needed to be checked etc

candleindark:
I modified `test.yml` to conform to the [documentation](https://docs.codecov.com/docs/adding-the-codecov-token) for adding a Codecov token to Github Actions. Please checkout the documentation, I think more is needed to be done in places I don't have the permission to access. In particular, according to the documentation, it seems that the token is specific to an individual repo and not to the entire organization.
Screenshot 2024-02-09 at 11 22 21 AM

PR #300, opened 2024-02-09
Provide script to auto restart Datalad-Registry upon host restart

### LINKED ISSUE #299: Auto-start whenever typhon is rebooted
For that create a script which would start the service and call it upon `@reboot`. E.g. this is how it looks for `shub` service:

```sh
shub@falkor:~$ crontab -l | grep -v '^#'
@reboot /srv/datasets.datalad.org/www/shub/_service_/service.sh
```

so you can create similar script and add a record for it via `crontab -e` command:

```sh
shub@falkor:~$ cat /srv/datasets.datalad.org/www/shub/_service_/service.sh
#!/bin/bash

set -eu
umask 077
cd $(dirname $0) && cd ..
singularity run --no-home -e -B /srv/datasets.datalad.org/shub/logs -B $PWD _service_/service.sif _service_/serve.py _data_/images.json &
disown
```

relates to #280 that you might want to reuse the "script" to start.

candleindark:
This PR adds a script to be used by cron to auto restart Datalad-Registry upon a rebooting of the host. It closes #299.

Note: The change in this PR has been applied to the live instances of DL-Registry.


tools/service.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# This script can be used by cron to auto restart Datalad-Registry service upon
+# the host's reboot.
+
+set -eu
+umask 077
+
+# Check if the path to the Datalad-Registry repo is provided
+if [ -z "$1" ]; then
```
yarikoptic, line 10: due to `set -eu` -- this would fail AFAIK... if it does, remove `set -eu` above or
```suggestion
if [ -z "${1:-}" ]; then
```

tools/service.sh
```diff
@@ -0,0 +1,18 @@
...
+set -eu
+umask 077
+
+# Check if the path to the Datalad-Registry repo is provided
+if [ -z "${1:-}" ]; then
+    echo "Usage: $0 <directory-path of Datalad-Registry repo>"
+    exit 1
+fi
+
+cd "$1"
```
yarikoptic, line 15: overall -- are you starting from this repo? if so -- why didn't you take the pattern I gave you in https://github.com/datalad/datalad-registry/issues/299 ie 

```
cd $(dirname $0) && cd ..
```
and thus avoiding needing `$1` altogether?

tools/service.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# This script can be used by cron to auto restart Datalad-Registry service upon
```
yarikoptic, line 3: it seems to "start", and not necessarily "restart", right?  just to make sure
```suggestion
# This script can be used by cron to auto start Datalad-Registry service upon
```

tools/service.sh
candleindark: No, I am not starting from this repo. I am starting from https://github.com/candleindark/datalad-registry, and depending on what latest features we want to include (also tests to an extend),  I start from a branch other than the master branch sometime.

candleindark: Wait, by repo. Do you mean the containing folder? 

I thought that cron's current working directory may not be the directory containing the script.

However, if cron's current working directory is not guarantee to be the directory containing the script, I thing I have to change the following line as well.

`podman-compose -f docker-compose.dev.yml --env-file .env.dev.typhon up -d --build`

Is cron guaranteed to have a current working directory that is the directory the script resides?

yarikoptic: > I thought that cron's current working directory may not be the directory containing the script.

correct.

> Is cron guaranteed to have a current working directory that is the directory the script resides?

not guaranteed.

That is why that `cd $(dirname $0)` I recommended -- to get to the directory containing the script! This way you do not need to provide long path to the script and then again that script as argument.

this all has nothing to do with branches.

candleindark: The script has been updated, and the `cd $(dirname $0) && cd ..` pattern is now used in the script.

PR #303, opened 2024-02-14
Configure the web UI so that the 'Nr of Annexed files' and 'Size of Annexed files' headers occupy different columns

### LINKED ISSUE #302: "Size of annexed files" is empty/null
regardless how I sort nothing pops up -- e.g. look at http://registry.datalad.org/overview/?filter=&sort=annexed_files_in_wt_size-desc

image

it does show number of annexed files, but not the size.

candleindark:
This PR configure the web UI so that the 'Nr of Annexed files' and 'Size of Annexed files' headers occupy different columns. 

This closes #302 which is due to table cells not being able to align to table headers.

### DIFF
--- a/datalad_registry/static/main.css
+++ b/datalad_registry/static/main.css
@@ -64,7 +64,6 @@ div#datalad-registry div.pagination {
 
 .tooltip {
     position: relative;
-    display: inline-block;
     border-bottom: 1px dotted black;
 }
 


datalad_registry/static/main.css
```diff
@@ -64,7 +64,6 @@ div#datalad-registry div.pagination {
 
 .tooltip {
     position: relative;
-    display: inline-block;
```
yarikoptic, line 67: ping @jsheunis  as we torturing your baby here to fix #302 - may be there is a better solution?

yarikoptic:
Let's proceed for now.

datalad_registry/static/main.css
yarikoptic: Can't give binary answer ;) -- I have not analyzed the cause of the bug, so can't attest that it happens only "where the file size column cells are empty" but "yes" -- after removing this added CSS it seems that it started to behave as desired.

Issue #301, opened 2024-02-13
Flower keeps busy IMHO unnecessarily in CPU and IO: needs further investigation


yarikoptic:
I think this needs further investigation since it remains the busiest process whenever there seems no activity otherwise.

CPU -- spikes into 100% averaging to ~60%

```shell
root@typhon:/home/yoh/proj/dandi/zarr-benchmarking/tools# pidstat -p 64122 5
Linux 6.1.0-17-amd64 (typhon)   02/13/2024      _x86_64_        (32 CPU)

09:22:33 AM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
09:22:38 AM  1036     64122   52.60    3.80    0.00    0.00   56.40     7  celery
09:22:43 AM  1036     64122   71.60    4.00    0.00    0.00   75.60     4  celery
09:22:48 AM  1036     64122   52.60    3.40    0.00    0.20   56.00    18  celery
09:22:53 AM  1036     64122   52.20    3.20    0.00    0.00   55.40     3  celery

root@typhon:/home/yoh/proj/dandi/zarr-benchmarking/tools# pidstat -p 64122 1
Linux 6.1.0-17-amd64 (typhon)   02/13/2024      _x86_64_        (32 CPU)

09:22:55 AM   UID       PID    %usr %system  %guest   %wait    %CPU   CPU  Command
09:22:56 AM  1036     64122   26.00    2.00    0.00    0.00   28.00     2  celery
09:22:57 AM  1036     64122   81.00    5.00    0.00    0.00   86.00    18  celery
09:22:58 AM  1036     64122   27.00    2.00    0.00    0.00   29.00     3  celery
09:22:59 AM  1036     64122  100.00    0.00    0.00    0.00  100.00     3  celery
09:23:00 AM  1036     64122   57.00    8.00    0.00    0.00   65.00     3  celery
09:23:01 AM  1036     64122   73.00    7.00    0.00    0.00   80.00     3  celery
09:23:02 AM  1036     64122   25.00    2.00    0.00    0.00   27.00    18  celery
```

and according to `iotop` it writes at about `15MBps` on average:

```
Total DISK READ:         0.00 B/s | Total DISK WRITE:        15.91 M/s
Current DISK READ:       0.00 B/s | Current DISK WRITE:      17.11 M/s
    TID  PRIO  USER     DISK READ DISK WRITE>    COMMAND                                                                                                                                                                                                           
[...]

PR #306, opened 2024-02-19
Documentation improvement

candleindark:
This PR achieves the following goals:

1. Brings the documentation up-to-date with the development
2. Replaces all the examples involving Docker with Podman (this PR closes #268)
3. Makes the flow in the documentation easier to follow
4. Ensures the illustrated procedures work in a clean Debian installation
5. Rename Dockerfile and Docker compose files so that development and production related files are easily distinguishable.

PR #307, opened 2024-02-19
Increase the interval between savings of `Flower`'s state

### LINKED ISSUE #258: celery (flower?) is "too busy" doing nothing/not much
Looking at typhon ATM it seems that celery processes are consuming considerable amount of CPU time but actually seems to be doing nothing useful since I don't see them spawning any underlying processes etc. I could be wrong, and it could be all Python-space metadata extraction I guess but it seems to be primarily the monitoring service (flower).  Worth researching into what it is doing.  Might be worth checking with `py-spy top` on what it actually is busy with etc.

candleindark:
This PR increases the time interval between savings of Flower's state. Each saving of Flower's state causes a complete re-write of the [DB](https://flower.readthedocs.io/en/latest/config.html#db) for storing Flower's state, so this interval must be sufficiently large to avoid overworking the CPU and the storage hardware.

This PR closes #258 and #301.

Note: Once this PR is merged. We need to switch the directory containing the DB file, `flower`, back to a spindle hard drive.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -85,7 +85,7 @@ services:
       FLOWER_BROKER_API: "${FLOWER_BROKER_API}"
       FLOWER_PERSISTENT: "True"
       FLOWER_DB: "/data/flower"
-      FLOWER_STATE_SAVE_INTERVAL: "60000"  # 1 minutes (in milliseconds)
+      FLOWER_STATE_SAVE_INTERVAL: "60000"  # 1 minute (in milliseconds)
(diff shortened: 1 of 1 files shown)


candleindark:
After one day of operation. The monitoring service is not consuming undue amount of CPU usage.

@yarikoptic I am merging this to the master now so that I can merge this change to https://github.com/datalad/datalad-registry/pull/306 in which I am modifying the Docker compose files. Additionally, we need arrange a time to to switch the directory containing the DB file, flower, back to a spindle hard drive.

PR #308, opened 2024-02-21
Add basic but flexible query language

yarikoptic:
An original target fake query which has various features was

```
((jim AND NOT haxby AND "important\" paper") OR ds_id:~"^000[3-9]..$" OR url:"example.com") AND metadata:non AND metadata[ex1,ex2]:"specific data" AND metadata[extractor2]:data
```
and this PR implements most of them -- check the modal window description of supported syntax:

image

- quite flexible implementation so we could add/replace fields we are searching on
- can search within specified metadata extractors
  - TODO (not here): search within JSON records is not part of this PR -- but likely would just allow to pass [JSONPath]( https://www.postgresql.org/docs/current/datatype-json.html#DATATYPE-JSONPATH) for the most flexible functionality
  - TODO (not here): expose via API call. #269 
- has help on syntax supporting boolean operations etc. Closes #264
- replacing original `/search`

PR #315, opened 2024-02-24
Handle `NULL` column values correctly in `ilike` queries

candleindark:
This PR addresses the issue describe in https://github.com/datalad/datalad-registry/pull/308#issuecomment-1961967932. The changes in this PR passes the tests describe in the issue. However, it causes some failures in other tests. I am unable to resolve the new failures at this point because they seems to be related to lark.

The rationale behind this fix as explained by ChatGPT.

> 
> You're correct in pointing out a misconception in my previous explanation regarding the evaluation of conditions involving `NULL` values in SQL. Let's clarify the behavior, especially in the context of your question about the behavior of `NULL` with `ILIKE` and logical operators.
> 
> In SQL, when you evaluate an expression involving `NULL` with a comparison operator (such as `=`, `!=`, `<`, `>`, `LIKE`, `ILIKE`, etc.), the result is `NULL`, which represents an unknown value. However, the logical treatment of `NULL` in conditional expressions like `WHERE` clauses behaves a bit differently when combined with logical operators (`AND`, `OR`, `NOT`).
> 
> ### The Correct Behavior
> 
> When you use an expression like:
> 
> ```sql
> (repo_url.url NOT ILIKE '%handbook%' OR repo_url.url IS NULL)
> ```
> 
> The evaluation logic works as follows:
> 
> - **`repo_url.url NOT ILIKE '%handbook%'`**: If `repo_url.url` is `NULL`, this condition evaluates to `NULL` because the result of any comparison with `NULL` is unknown (`NULL`).
> 
> - **`repo_url.url IS NULL`**: This explicitly checks for `NULL` and evaluates to `true` if `repo_url.url` is indeed `NULL`.
> 
> When combined with `OR`:
> 
> - If `repo_url.url` is `NULL`, the left side of the `OR` (`repo_url.url NOT ILIKE '%handbook%'`) evaluates to `NULL`, but the right side (`repo_url.url IS NULL`) evaluates to `true`. Since `NULL OR true` is `true` in SQL's logical operations, the overall condition will evaluate to `true`, thus including the row in the results.
> 
> In contrast to my incorrect implication before, the presence of `OR repo_url.url IS NULL` ensures that rows where `repo_url.url` is `NULL` are indeed considered and not excluded based on the `ILIKE` comparison alone. The logical `OR` ensures that if either condition is `true` (or in SQL terms, if either side evaluates to `true` or `NULL` and the other to `true`), the overall expression is considered `true`.
> 
> So, for a row where `repo_url.url` is `NULL`, the evaluation of `(repo_url.url NOT ILIKE '%handbook%' OR repo_url.url IS NULL)` effectively becomes `true`, ensuring such rows are included in the query results, contrary to the concern of it evaluating to `NULL` and being excluded.
> 
> I appreciate your patience and the opportunity to correct this explanation. Your understanding of the behavior of `NULL` in SQL is indeed correct, and it's an important aspect of SQL query formulation to grasp.

Additionally,

> **In conditional expressions**, such as those in `WHERE` clauses, a condition that evaluates to `NULL` does not qualify a row for inclusion in the result set. This is because SQL requires conditions to be explicitly `true` to include a row, and `NULL` is considered neither `true` nor `false`.

### DIFF
--- a/datalad_registry/search.py
+++ b/datalad_registry/search.py
@@ -53,7 +53,7 @@ def get_branches_ilike_search(value):
 known_fields = {
     _: partial(get_ilike_search, RepoUrl, _) for _ in known_fields_RepoUrl_1to1
 }
-known_fields["metadata"] = get_metadata_ilike_search  # type: ignore
+
 known_fields["branches"] = get_branches_ilike_search  # type: ignore
 
 # TODO: add "metadata_extractor"?
@@ -216,7 +216,13 @@ class SearchQueryTransformer(Transformer):
 
     def _get_str_search(self, arg: Token) -> ColumnElement[bool]:
         value = self._get_str_value(arg)
-        return or_(*[f(value) for f in known_fields.values()])
+        return or_(
+            *(
+                and_(getattr(RepoUrl, field).is_not(None), ilike_query_gen(value))
+                for field, ilike_query_gen in known_fields.items()
+            ),
+            get_metadata_ilike_search(value),
+        )
 
     search_word = _get_str_search
     search_string = _get_str_search
--- a/datalad_registry/search.py
+++ b/datalad_registry/search.py
@@ -30,7 +30,11 @@ known_fields_RepoUrl_1to1 = ["url", "ds_id", "head", "head_describe", "tags"]
 
 
 def get_ilike_search(model, field: str, value: str):
-    return getattr(model, field).ilike(_escape_for_ilike(value), escape=escape)
+    model_field = getattr(model, field)
+    return and_(
+        model_field.is_not(None),
+        model_field.ilike(_escape_for_ilike(value), escape=escape),
+    )
 
 
 def get_metadata_ilike_search(value):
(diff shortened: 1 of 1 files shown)


datalad_registry/search.py
```diff
@@ -53,7 +53,7 @@ def get_branches_ilike_search(value):
 known_fields = {
     _: partial(get_ilike_search, RepoUrl, _) for _ in known_fields_RepoUrl_1to1
 }
-known_fields["metadata"] = get_metadata_ilike_search  # type: ignore
+
```
yarikoptic, line 56: This removes search within 'metadata' field entirely, which definitely would cause failures

candleindark: Yeah, I just found that out as well. I have fixed that, so all tests passe at this point.

candleindark:
I added the `"metadata"` field back to `known_fields`. However,  I added it after the setting of `known_fields["branches"]`. This way, the generated SQL looks more clean though I can't say if it is more "correct".

yarikoptic:
Woohoo!!!! Thank you for digging the fix!!!

Issue #224, opened 2023-08-21
Make testing docker-compose setup "transparent" to the developer


yarikoptic:
previously mentioned in #36

In dandi-cli we have setup a session level fixture which brings up necessary docker compose recipe up and down, and also has ways (environment variables) to persist or force update upon start: https://github.com/dandi/dandi-cli/blob/HEAD/dandi/tests/fixtures.py#L372 

Adding such a fixture
- avoids need to manually establish env variables (you can just do what needs to be done in that fixture for the env vars)
- streamlines for a developer running the tests
- removes similar boiler plate from CI workflows (in `.github/workflows/test.yml`)

note:
- files needed for fixture would need to be moved under `datalad_registry/` folder somewhere to be shipped within source distribution along with the test, e.g. how in dandi-cli: https://github.com/dandi/dandi-cli/blob/HEAD/dandi/tests/data/dandiarchive-docker/docker-compose.yml

yarikoptic:
@candleindark please update me on where are we with this.

candleindark:
> @candleindark please update me on where are we with this.

I have not worked on this yet. I am still working on the update feature, but I am getting close to finishing that one though.

yarikoptic:
@jwodder in 
- https://github.com/datalad/datalad-registry/pull/317#discussion_r1507905742
reminded about https://github.com/theskumar/python-dotenv which he used in e.g.
- https://github.com/con/tinuous/blob/9ca71bcc2e4b6129c9e37629014a0d350a033669/src/tinuous/__main__.py#L12
which might come handy to automate picking up the container env vars setup before starting up the `podman-compose up` in the fixture.

PR #317, opened 2024-02-26
Allow testing while running DL-Registry in `DEVELOPMENT` mode in the same machine

### LINKED ISSUE #314: Ensure testing can be done while a development instance of Datalad-Registry is running
Modify the needed Docker compose files and .env file so that testing can be done while a development instance of Datalad-Registry is running

candleindark:
The goal of this PR is to allow the running of the tests while running DL-Registry in DEVELOPMENT mode in the same machine. Thus, this PR closes #314. Additionally, this PR renames some of the test configuration files and improves the `test` GitHub workflow.  `docker-compose.testing.yml` is renamed to `docker-compose.test.yml` and `env.testing` to `env.test`. The new names are less verbose. The `test` GitHub workflow now loads environment variables directly from `env.test` instead of relying on re-specification.

In achieving the main goal, this PR

1. assigns different ports to the test components of DL-Registry to avoid port conflicts with a running instance of DL-Registry in `DEVELOPMENT` mode.
2. sets Podman/Docker project name to `dl-registry-test` so that the test components have a dedicated virtual network.
3. Configure pytest to ignore the default directories created by an instance of DL-Registry in `DEVELOPMENT` mode.


.github/workflows/test.yml
```diff
@@ -47,12 +37,14 @@ jobs:
...
+      run: rm -rf build dist
     - name: Start services with Docker Compose
-      run: docker compose -f docker-compose.testing.yml up -d --build
+      run: (set -a && . ./env.test && set +a && docker compose -f docker-compose.test.yml up -d)
     - name: Run tests with Coverage
-      run: pytest --cov=. --cov-report=xml
+      run: (set -a && . ./env.test && set +a && python -m pytest -s -v --cov=. --cov-report=xml)
     - name: Stop services provided by Docker Compose
-      run: docker compose -f docker-compose.testing.yml down
+      run: (set -a && . ./env.test && set +a && docker compose -f docker-compose.test.yml down)
```
yarikoptic, line 47: @jwodder is there a more convenient way to inherit changes in environment in one step across all steps?   pretty much ./env.test has variables which we need to export/use in those steps

jwodder: @yarikoptic If all the lines in `env.test` are of the form `varname=value`, you can append it to [the file in `$GITHUB_ENV`](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable).

candleindark: One reason I want to do it this way is that it is more inline with the instruction for setting up testing at the [README](https://github.com/datalad/datalad-registry?tab=readme-ov-file#to-run-tests).  The only difference between this line and the one in the README instruction is the use of `docker compose` instead of `podman-compose`.

This arrangement can be better at helping us to pick up some inconsistencies between our code and documentation.

yarikoptic: I can see how it could be beneficial, but my allergy to duplication and clutter which then makes it harder to find defects almost outweighs a possible benefit from a distant harmonization with the README, since in effect it would be achieving the same.

If you were to test "instructions in README", here is an example how I do it https://github.com/ReproNim/containers/blob/master/.github/workflows/tests.yaml#L30 -- I really take and run it.  Duplicating (duplication is evil) instruction just opens the way for them to diverge eventually.

So overall  - even if motivation is sound, long term IMHO it is not beneficial and might be confusing for someone looking at it ("why they didn't just `cat env.test >>  "$GITHUB_ENV"`?) and  it is **already** different from README (docker vs podman).

.github/workflows/test.yml
```diff
@@ -47,12 +37,38 @@ jobs:
...
+            # Skip empty lines and lines starting with '#'
+            if [[ -z "$trimmed_line" || "$trimmed_line" =~ ^# ]]; then
+                continue
+            fi
+            # Extract variable name (part before the first '=')
+            variable_name=$(echo "$trimmed_line" | cut -d '=' -f 1)
+
+            # Set the needed env vars in the GitHub Actions workflow
+            echo "$variable_name=${!variable_name}" >> "$GITHUB_ENV"
+        done < "$env_file"
```
yarikoptic, line 65: Why such complication is necessary or why 

```
cat env.test >> "$GITHUB_ENV"
```
is insufficient? ie. have you tried it and it has failed how?

candleindark: No, doing that isn't sufficient.

Here are the reasons.

1. GH workflow doesn't do variable expansion, which is needed in defining some of the variables in `env.test`. I tried a verified this behavior of GH workflow yesterday.
2. According to John's earlier post, all the lines appended to `"$GITHUB_ENV"` must be of the form `varname=value`. Since the `env.test` file contains comments, so appending the content of `env.test` without modification to `"$GITHUB_ENV"` wouldn't work. I just verified this behavior with https://github.com/datalad/datalad-registry/pull/317/commits/c9e52d95c0cf07e8a7dd382957e38a4f4fbd9137 and https://github.com/datalad/datalad-registry/pull/317/commits/38fa0e96b2dfc162d7eb7099cefaf56dccae2e16.

yarikoptic: ah, I was not aware that it cannot do variables substitution right there -- that's kinda odd... let me think about that.
As for comments -- trivial to address with smth like

```
cat env.testing | sed -e 's, *#.*,,g' | grep -v '^$' >> "$GITHUB_ENV"
```

getting back to above aspect of expansions, with chatgpt I could only master following abomination

```
❯ /bin/bash -c 'declare -p | sort > /tmp/before.env; source env.testing ; declare -p | sort > /tmp/after.env; comm -13 /tmp/before.env /tmp/after.env | grep -v -e "\(_=\|LINES\|COLUMNS\)" | sed -ne "/--/s,.*-- *,,gp"'
CELERY_BROKER_URL="amqp://tester:testpass@localhost:5672"
CELERY_RESULT_BACKEND="redis://localhost:6379"
POSTGRES_DB="pgdb"
POSTGRES_PASSWORD="testpass"
POSTGRES_USER="tester"
RABBITMQ_DEFAULT_PASS="testpass"
RABBITMQ_DEFAULT_USER="tester"
SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://tester:testpass@localhost:5432/pgdb"
```
which I also dislike... 

@jwodder -- do you know/see any pretty way to inherit changes in environment from `. env.testing` across steps or instruct github susbtitute variable names within GITHUB_ENV?

jwodder: @yarikoptic You might be able to use [`python-dotenv`](https://github.com/theskumar/python-dotenv)'s CLI to process `env.testing` into just variable assignments with expansions applied.

yarikoptic: thanks! added  a note on that to
- https://github.com/datalad/datalad-registry/issues/224#issuecomment-1971688641

for here/now I think we can simplify it a little

```shell
env_file=./env.testing

set -a
. "$env_file"
set +a

sed -e "s, *#.*,,g" < "$env_file" \
| grep -v '^$' \
| while IFS="=" read -r var _; do 
    echo $var="${!var}"; 
done;
```

which produces
```shell
❯ bash /tmp/1.sh
RABBITMQ_DEFAULT_USER=tester
RABBITMQ_DEFAULT_PASS=testpass
POSTGRES_DB=pgdb
POSTGRES_USER=tester
POSTGRES_PASSWORD=testpass
CELERY_BROKER_URL=amqp://tester:testpass@localhost:5672
CELERY_RESULT_BACKEND=redis://localhost:6379
SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://tester:testpass@localhost:5432/pgdb
```

candleindark: > ```shell
> env_file=./env.testing
> 
> set -a
> . "$env_file"
> set +a
> 
> sed -e "s, *#.*,,g" < "$env_file" \
> | grep -v '^$' \
> | while IFS="=" read -r var _; do 
>     echo $var="${!var}"; 
> done;
> ```

That's great. However, since we are going to use python-dotenv in our project for other purposes as well (good idea). We can leverage the CLI of python-dotenv as @jwodder recommended earlier to simplify the `test.yml` file even further (see my latest commits).

yarikoptic: oh cool -- I guess I have missed that it could be used that way!

yarikoptic:
test instance works for me but the dev doesn't work with podman even on master, so filed a separate issue:
- https://github.com/datalad/datalad-registry/issues/322
and meanwhile let's proceed on this one

PR #324, opened 2024-03-08
Expose search functionality in API

### LINKED ISSUE #269: Expose current search via API
There is already interest (if not need) in using registry and its API (attn @ericearl and @remi-gau)
We have the `/api/v2/dataset-urls` endpoint which is kinda "too heavy" with already ways to limit the list of datasets.  But I guess could be loaded with one more exposing that basic search we have in the edit box.  That would also be useful for #163 . An alternative is to add `/api/v2/search` endpoint which would be similar to `/dataset-urls` one in return  but just limit based on search.

candleindark:
This PR exposes the search functionality available in the web UI through the API. It closes #269.

The search functionality is exposed via the query parameter `search` to the existing collection endpoint of `/api/v2/dataset-urls`. This parameter can be mix with other parameters at that endpoint to create more restricted searchs.

Note: The changes in this PR has been applied to the live instances of DL-Registry.

### DIFF
--- a/datalad_registry/blueprints/api/dataset_urls/__init__.py
+++ b/datalad_registry/blueprints/api/dataset_urls/__init__.py
@@ -8,9 +8,8 @@ from celery import group
 from flask import current_app, url_for
 from flask_openapi3 import APIBlueprint, Tag
 from psycopg2.errors import UniqueViolation
-from sqlalchemy import and_
+from sqlalchemy import ColumnElement, and_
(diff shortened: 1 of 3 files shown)


yarikoptic:
is there much more to it besides adding a test case or two?

candleindark:
> is there much more to it besides adding a test case or two?

No. The tests are added.

yarikoptic:
Great, thanks!

PR #344, opened 2024-03-24
Restrict inputs to `search.parse_query` to non-empty strings

candleindark:
This PR restricts inputs to `search.parse_query` to non-empty strings. An empty query string as an input to the parser will result in `lark.exceptions.UnexpectedEOF` anyway. It is more explicit and informative to rule out this input in the `parse_query` function level.


yarikoptic:
Fine with me, thank you. Let's proceed

PR #353, opened 2024-04-12
Provide stats about return collection

### LINKED ISSUE #349: Provide "total stats"
Following up on @joeyh question -- we should do the SUM on the "Size of annexed files" and "Nr of annexed files" first taking max per each ds_id to get a total sum across different dataset IDs.

- [x] Provide the stats through the API
- [ ] Provide the stats through the web UI

candleindark:
This PR provides stats regarding the returned collection of a query of dataset URLs. It closes #349.

Stats on unique datasets are only calculated on Datalad datasets. We can extend the changes in this PR to provide stats on guessed unique non-Datalad datasets in the future harvesting more info from the repos and choosing the logic that does the guessing work.

Note: 

- The changes in this PR has been applied to the instance on Typhon but not yet Falkor.
- With stats gathered, the search function is palpably slower.

### DIFF
--- a/datalad_registry/blueprints/api/dataset_urls/tools.py
+++ b/datalad_registry/blueprints/api/dataset_urls/tools.py
@@ -1,4 +1,16 @@
-from sqlalchemy import Select, Subquery, and_, func, not_, or_, select
+from sqlalchemy import (
+    Select,
+    Subquery,
+    TableClause,
+    and_,
+    column,
+    func,
+    not_,
+    or_,
+    select,
+    table,
+    text,
+)
 
 from datalad_registry.models import RepoUrl, db
 
@@ -9,6 +21,29 @@ from .models import (
     NonAnnexDsCollectionStats,
     StatsSummary,
 )
+
+
+def cache_result_to_tmp_tb(select_stmt: Select, tb_name: str) -> TableClause:
+    """
+    Execute the given select statement and cache the result to a temporary table
+    with the given name
+
+    :param select_stmt: The given select statement to execute
+    :param tb_name: The string to use as the name of the temporary table
+    :return: A object representing the temporary table
+
+    Note: The execution of this function requires the Flask app's context
+    """
+    create_tmp_tb_sql = f"""
+        CREATE TEMPORARY TABLE {tb_name} AS
+        {select_stmt.compile(bind=db.engine, compile_kwargs={'literal_binds': True})};
+    """
+    db.session.execute(text(create_tmp_tb_sql))
+
+    return table(
+        tb_name,
+        *(column(name, c.type) for name, c in select_stmt.selected_columns.items()),
+    )
 
 
 def _get_annex_ds_collection_stats(q: Subquery) -> AnnexDsCollectionStats:
--- a/datalad_registry/blueprints/api/dataset_urls/tools.py
+++ b/datalad_registry/blueprints/api/dataset_urls/tools.py
@@ -1,0 +1,180 @@
+from sqlalchemy import Select, Subquery, and_, func, or_, select
+
+from datalad_registry.models import RepoUrl, db
+
+from .models import (
+    AnnexDsCollectionStats,
+    CollectionStats,
+    DataladDsCollectionStats,
+    NonAnnexDsCollectionStats,
+    StatsSummary,
+)
+
+
+def _get_annex_ds_collection_stats(q: Subquery) -> AnnexDsCollectionStats:
+    """
+    Get the stats of a collection of datasets that contains only of annex datasets
+
+    :param q: The query that specifies the collection of datasets under consideration
+    :return: The object representing the stats
+
+    Note: The execution of this function requires the Flask app's context
+    """
+
+    ds_count, annexed_files_size, annexed_file_count = db.session.execute(
+        select(
+            func.count().label("ds_count"),
+            func.sum(q.c.annexed_files_in_wt_size).label("annexed_files_size"),
+            func.sum(q.c.annexed_files_in_wt_count).label("annexed_file_count"),
+        ).select_from(q)
+    ).one()
+
+    return AnnexDsCollectionStats(
+        ds_count=ds_count,
+        annexed_files_size=annexed_files_size,
+        annexed_file_count=annexed_file_count,
+    )
+
+
+def get_unique_dl_ds_collection_stats(base_q: Subquery) -> AnnexDsCollectionStats:
+    """
+    Get the stats of the subset of the collection of datasets that contains only
+    of Datalad datasets, considering datasets with the same `ds_id` as the same
+    dataset
+
+    :param base_q: The base query that specified the collection of datasets
+                   under consideration
+    :return: The object representing the stats
+
+    Note: The execution of this function requires the Flask app's context
+    """
+
+    grp_by_id_q = (
+        select(
+            base_q.c.ds_id,
+            func.max(base_q.c.annexed_files_in_wt_size).label(
+                "max_annexed_files_in_wt_size"
+            ),
+        )
+        .group_by(base_q.c.ds_id)
+        .subquery("grp_by_id_q")
+    )
+
+    grp_by_id_and_a_f_size_q = (
+        select(
+            RepoUrl.ds_id,
+            RepoUrl.annexed_files_in_wt_size,
+            func.max(RepoUrl.annexed_files_in_wt_count).label(
+                "annexed_files_in_wt_count"
+            ),
+        )
+        .join(
+            grp_by_id_q,
+            and_(
+                RepoUrl.ds_id == grp_by_id_q.c.ds_id,
+                or_(
+                    grp_by_id_q.c.max_annexed_files_in_wt_size.is_(None),
+                    RepoUrl.annexed_files_in_wt_size
+                    == grp_by_id_q.c.max_annexed_files_in_wt_size,
+                ),
+            ),
+        )
+        .group_by(RepoUrl.ds_id, RepoUrl.annexed_files_in_wt_size)
+        .subquery("grp_by_id_and_a_f_size_q")
+    )
+
+    return _get_annex_ds_collection_stats(grp_by_id_and_a_f_size_q)
+
+
+def get_dl_ds_collection_stats_with_dups(base_q: Subquery) -> AnnexDsCollectionStats:
+    """
+    Get the stats of the subset of the collection of datasets that contains only
+    of Datalad datasets, considering individual repos as a dataset regardless of
+    the value of `ds_id`.
+
+    :param base_q: The base query that specified the collection of datasets
+                   under consideration
+    :return: The object representing the stats
+
+    Note: The execution of this function requires the Flask app's context
+    """
+
+    # Select statement for getting all the Datalad datasets
+    dl_ds_q = select(base_q).filter(base_q.c.ds_id.is_not(None)).subquery("dl_ds_q")
+
+    return _get_annex_ds_collection_stats(dl_ds_q)
+
+
+def get_dl_ds_collection_stats(base_q: Subquery) -> DataladDsCollectionStats:
+    """
+    Get the stats of the subset of the collection of datasets that contains only
+    of Datalad datasets
+
+    :param base_q: The base query that specified the collection of datasets
+                   under consideration
+    :return: The object representing the stats
+
+    Note: The execution of this function requires the Flask app's context
+    """
+
+    return DataladDsCollectionStats(
+        unique_ds_stats=get_unique_dl_ds_collection_stats(base_q),
+        stats=get_dl_ds_collection_stats_with_dups(base_q),
+    )
+
+
+def get_pure_annex_ds_collection_stats() -> AnnexDsCollectionStats:
+    """
+    Get the stats of the subset of the collection of datasets that contains only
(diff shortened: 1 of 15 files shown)


candleindark:
@yarikoptic What is the definitive way to determine if a git repo is a git annex repo? It seems that I can't decide if a repo is a git annex repo based on whether it has an `annex-uuid`. I found some repos without an `annex-uuid` that have `annexed_files_in_wt_count` and `annexed_files_size`.

yarikoptic:
> What is the definitive way to determine if a git repo is a git annex repo?

in general -- e.g. if it has `git-annex` branch.  But if you are asking if it is a git annex repo which provides storage for annexed files (e.g. repo on github is **not** such), then it is the ones for which git-annex didn't set `annex-ignore`.

candleindark:
In #349, you want DL-Registry to provide stats about pure annex repos (annex repos that are not Datalad datasets). Given that as the context, how do you want a repo be considered a git annex repo, having the `git-annex` branch? 

Check whether a repo has a `git-annex` branch can be done through the [`RepoUrl` model](https://github.com/datalad/datalad-registry/blob/4da200b67004c034148944713ee445031f31989e/datalad_registry/models.py#L15-L82). If the decision relies on information that's not currently provided by the `RepoUrl` model, we have to update the model further process the repos in our local cache/storage.

yarikoptic:
> how do you want a repo be considered a git annex repo, having the `git-annex` branch?

yes -- should be enough.  But we are yet to figure out some way to identify "unique" collection of such repos since there is no uuid for them like in the case of datalad datasets.

candleindark:
> But we are yet to figure out some way to identify "unique" collection of such repos since there is no uuid for them like in the case of datalad datasets.

Is it true that the best we can do in identifying `unique` non-Datalad repos is by guessing, and the result will be an approximation?

yarikoptic:
AFAIK yes -- there is no *explicit/precense* way.

yarikoptic:
```
FAILED datalad_registry/tests/test_blueprints/test_api/test_dataset_urls.py::TestDatasetURLs::test_valid_query_params[query_params26] - pydantic.error_wrappers.ValidationError: 2 validation errors for AnnexDsCollectionStats
annexed_files_size
  none is not an allowed value (type=type_error.none.not_allowed)
annexed_file_count
  none is not an allowed value (type=type_error.none.not_allowed)
```

candleindark:
> ```
> FAILED datalad_registry/tests/test_blueprints/test_api/test_dataset_urls.py::TestDatasetURLs::test_valid_query_params[query_params26] - pydantic.error_wrappers.ValidationError: 2 validation errors for AnnexDsCollectionStats
> annexed_files_size
>   none is not an allowed value (type=type_error.none.not_allowed)
> annexed_file_count
>   none is not an allowed value (type=type_error.none.not_allowed)
> ```


Fixed per https://github.com/datalad/datalad-registry/pull/353/commits/926d8352a5d8d0218756f9de202cb781cb5236bc.  Thanks for pointing out.

datalad_registry/templates/overview.html
```diff
@@ -178,35 +191,89 @@ <h2>Search query syntax</h2>
     <div class="pager">
       {{ render_pagination_widget(pagination, '.overview') }}
     </div>
+
+    <!-- Stats Trigger Button -->
+    <button id="ShowStats">Show Stats</button>
```
yarikoptic, line 196: so they aren't shown by default -- is that because it needs notable backend/DB effort to get them or for another reason?

yarikoptic: could you may be show screenshots of what it would look like?

candleindark: > so they aren't shown by default -- is that because it needs notable backend/DB effort to get them or for another reason?

That behavior is result of a solely stylistic choice, which maybe be a very good choice in the hindsight.  At https://github.com/datalad/datalad-registry/pull/353/commits/69490edd64020c7c5be948e442906875e9550b3f, the stats are printed directly at the end of the overview page.

candleindark: > could you may be show screenshots of what it would look like?

The changes had been applied to the Typhon instance yesterday, so you can try out the stats feature by accessing the web service on the Typhon instance.  Here are the screenshots though.


Screenshot 2024-04-20 at 1 01 01 PM
Screenshot 2024-04-20 at 1 01 24 PM

candleindark:
I think we can improve the performance of gathering stats by looking into some "caching" mechanism for the queried results.

datalad_registry/templates/overview.html
yarikoptic: Quick one: please use humanize for the displayed information (in API just size).
I think with showing which summary sentence with total size and number of files, and then adding button "show details" with this modal

yarikoptic: Also for me modal doesn't render correctly at 100%:

image

yarikoptic:
> I think we can improve the performance of gathering stats by looking into some "caching" mechanism for the queried results.

Do you have any estimates on how slower it has become?
If there is a notable slowdown, we might want to make it then optional: add a checkbox `[ ] Collect statistics` and if not checked - do query without stats.  I thought that it was not really effecting the query much.

candleindark:
> > I think we can improve the performance of gathering stats by looking into some "caching" mechanism for the queried results.
> 
> Do you have any estimates on how slower it has become? If there is a notable slowdown, we might want to make it then optional: add a checkbox `[ ] Collect statistics` and if not checked - do query without stats. I thought that it was not really effecting the query much.

I know what with stats is multiple times slower than without stats. I will look into how to `cache` query first. If that solution doesn't workout well, I will try the checkbox solution.

datalad_registry/templates/overview.html
candleindark: Fixed with https://github.com/datalad/datalad-registry/pull/353/commits/000a30d1f13fba67b5326ace1de9ec425f022055.

candleindark:
Most of the slow down is due to the fact that the search/filter query is being executed repeatedly as a subquery in the calculation of the stats. One can verify this by going through the pages without any search query (there is no much slow down even with the stats compiled).

I am working on a solution for this with creating a temporary table now.

candleindark:
@yarikoptic A temporary table is now used to `cache` the search/filter result so that in calculating the stats the search query doesn't have to be repeatedly executed. Before this change is put in place, the search query was executed 6 times for generating the result page with stats. Now it is executed only 2 times. Two times are need now because a temporary table doesn't retain the same relationship as the original table, so I need to query the original `repo_url` table to provide info on metadata on the result page, and separately do a query on to `repo_url` table to export the result to the temporary table on which all the stats are calculated. Over all, the performance is significantly improved.

The latest change has been pushed to Typhon. You can test the performance improvement by comparing the Typhon instance (which has the stats feature) with the Falkor instance (which doesn't have the stats feature).

Let me know if you still want me to implement the checkbox feature, https://github.com/datalad/datalad-registry/pull/353#issuecomment-2071159095, after this performance improvement.

datalad_registry/templates/overview.html
```diff
@@ -178,35 +191,99 @@ <h2>Search query syntax</h2>
     <div class="pager">
       {{ render_pagination_widget(pagination, '.overview') }}
     </div>
+
+
+    <div>
+      <span style="margin-right: 10px">Combined Nr of Annexed
```
yarikoptic, line 197:
```suggestion
      <span style="margin-right: 10px"># of annexed
```

datalad_registry/templates/overview.html
```diff
@@ -178,35 +191,99 @@ <h2>Search query syntax</h2>
...
+      files: {{ ((stats.datalad_ds_stats.stats.annexed_file_count if stats.datalad_ds_stats.stats.annexed_file_count else 0) +
+        (stats.pure_annex_ds_stats.annexed_file_count if stats.pure_annex_ds_stats.annexed_file_count else 0))|intcomma }}</span>
+      <span style="margin-right: 5px">Combined Annexed file
+        size: {{ ((stats.datalad_ds_stats.stats.annexed_files_size if stats.datalad_ds_stats.stats.annexed_files_size else 0) + (stats.pure_annex_ds_stats.annexed_files_size if stats.pure_annex_ds_stats.annexed_files_size else 0))|filesizeformat }}</span>
```
yarikoptic, line 201:
```suggestion
      <span style="margin-right: 5px">Combined size of annexed
        files: {{ ((stats.datalad_ds_stats.stats.annexed_files_size if stats.datalad_ds_stats.stats.annexed_files_size else 0) + (stats.pure_annex_ds_stats.annexed_files_size if stats.pure_annex_ds_stats.annexed_files_size else 0))|filesizeformat }}</span>
```

datalad_registry/templates/overview.html
```diff
@@ -178,35 +191,99 @@ <h2>Search query syntax</h2>
...
+
+      <!-- Stats Trigger Button -->
+      <button id="ShowStats">Show Details</button>
```
yarikoptic, line 204:
```suggestion
      <button id="ShowStats">Show details</button>
```

datalad_registry/templates/overview.html
```diff
@@ -178,35 +191,99 @@ <h2>Search query syntax</h2>
...
+          <span id="CloseStats" class="close">&times;</span>
+          <h2>Stats</h2>
+        </div>
+        <div id="stats" class="modal-body">
+          <ul>
+            <li>
+              <h3>Datalad Datasets Stats</h3>
+              <ul>
+                <li>
+                  <h4>Unique Datalad Dataset Stats</h4>
```
yarikoptic, line 221:
```suggestion
                  <h4>Unique DataLad dataset stats</h4>
```

datalad_registry/templates/overview.html
```diff
@@ -178,35 +191,99 @@ <h2>Search query syntax</h2>
...
+    <div id="StatsModal" class="modal">
+      <div class="modal-content">
+        <div class="modal-header">
+          <span id="CloseStats" class="close">&times;</span>
+          <h2>Stats</h2>
+        </div>
+        <div id="stats" class="modal-body">
+          <ul>
+            <li>
+              <h3>Datalad Datasets Stats</h3>
```
yarikoptic, line 218:
```suggestion
              <h3>DataLad datasets stats</h3>
```

datalad_registry/templates/overview.html
```diff
@@ -178,35 +191,99 @@ <h2>Search query syntax</h2>
...
+                  {{ render_annex_ds_collection_stats(stats.datalad_ds_stats.unique_ds_stats) }}
+                </li>
+                <li>
+                  <h4>Stats without deduplication</h4>
+                  {{ render_annex_ds_collection_stats(stats.datalad_ds_stats.stats) }}
+                </li>
+              </ul>
+            </li>
+            <li>
+              <h3>Pure Annex Dataset Stats</h3>
```
yarikoptic, line 231:
```suggestion
              <h3>Pure annex repositories stats</h3>
```

datalad_registry/templates/overview.html
```diff
@@ -178,35 +191,99 @@ <h2>Search query syntax</h2>
...
+              {{ render_annex_ds_collection_stats(stats.pure_annex_ds_stats) }}
+            </li>
+            <li>
+              <h3>Non-Annex Dataset Stats</h3>
```
yarikoptic, line 235:
```suggestion
              <h3>Non-annex repositories stats</h3>
```

are those datalad datasets or just repositories or either?

datalad_registry/blueprints/api/dataset_urls/tools.py
```diff
@@ -0,0 +1,238 @@
...
+    statement
+
+    :param select_stmt: The given select statement
+    :return: The statistics of the collection of dataset URLs
+
+    Note: The execution of this function requires the Flask app's context
+    """
+
+    # Cache the result of the select statement to a temporary table
+    tmp_tb = cache_result_to_tmp_tb(select_stmt, "tmp_tb")
```
yarikoptic, line 219: what removes this temporary table?  couldn't we remove it upon `return`? then make it `try/finally` structure with removing it in `finally`.

What would happen if two parallel requests come -- wouldn't they collide to create the same `tmp_tb` -- shouldn't name be random?

yarikoptic (changes requested):
- some minor wording in stats tune up (no need to Capitalize Every Word0
- I think there should care be taken about avoiding collisions in that temporary table and removing it after

datalad_registry/blueprints/api/dataset_urls/tools.py
yarikoptic: ok then please just attend to suggested tune up of text and let's be done with this one

datalad_registry/templates/overview.html
candleindark: They are git repos that don't have a `"git-annex"` branch. Referring them as repos are more precise, but in our code base, we loosely just call all repos datasets. The current arrangement in our code base is a bit confusing.

candleindark:
@yarikoptic The requested changes have been put in, and the latest changes in this PR and #360 have been applied to the Typhon instance. Please have a final review and merge this PR once you are satisfied. Once this PR is merge, I will make the changes public on the public instance.

yarikoptic:
Looks good. @joeyh -- so we have only slightly over 816TB covered so far total with some account for duplicates(clones):

image

PR #363, opened 2024-04-30
Implement `dandi:files` extractor

candleindark:
This PR implements the `dandi:files` extractor. Additionally, it reorganizes the Docker compose files and dot env template files so that port of services at host can be changed easily. This ease of change of service ports at host facilitate running different instances of DL-Registry in the same host, allowing experimentation of different setup within a host.


candleindark:
@yarikoptic This PR is ready for merge. Upon your decision on whether to make the extracted metadata by `dandi:files` available on the production instance of DL-Registry on Typhon, we can merge this PR with https://github.com/datalad/datalad-registry/blob/b6e9d5b5588786a4a53e9b64d6d5b2c818df4227/datalad_registry/conf.py#L56 commented or not.

datalad_registry/conf.py
```diff
@@ -53,7 +53,7 @@ class BaseConfig(OperationConfig):
         "bids_dataset",
         # === DANDI related extractors ===
         "dandi",
-        # "dandi:files",  # Let's not activate this yet by default
+        "dandi:files",
```
yarikoptic, line 56: for production we would not want it in this shape.

So we will just need to comment out here.

candleindark:
@yarikoptic I think we should merge this one if you are OK with the implementation. If you are concerned with the performance impact, we can run the registry with the `dandi:files` metadata extractor disabled (the default, see last comment). Beside implementing the `dandi:files` metadata extractor, this PR reorganizes the Docker compose files and dot env template files so that port of services at host can be changed easily. I want those changes available in the master branch.

These is a live instance of DL-Registry with the changes in this PR and the `dandi:files` metadata extractor enabled on Typhon. You can experience the changes of this PR though that instance.

PR #366, opened 2024-05-03
Refactor GitHub Actions workflows

candleindark:
This PR refactor GitHub Actions workflows in the following manner.

1. Rename some of the jobs to avoid confusion
2. Remove static type checking from the `test.yml` workflow to put it in a specific workflow

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,6 @@ jobs:
           pip install --upgrade pip
           pip install -r requirements.txt
           pip install wheel
-          pip install mypy types-requests types-PyYAML
           pip install .[test]
       - name: Remove build artifacts
         run: rm -rf build dist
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -58,8 +57,6 @@ jobs:
         if: ${{ github.event_name != 'workflow_dispatch' || !inputs.debug_enabled }}
       - name: Stop services provided by Docker Compose
         run: docker compose -f docker-compose.test.yml down
-      - name: mypy check
-        run: mypy datalad_registry datalad_registry_client
```
yarikoptic, line 62: to reuse conditions etc on the top for running testing, might be better to just create `mypy` job in addition to `test` job within the same file.

candleindark: Actually, the `mypy` run defined in the [`type_check.yml`](https://github.com/datalad/datalad-registry/blob/68e3d459edcdda51d5836527e9dfef325c98cde6/.github/workflows/type_check.yml) now has different [trigger](https://github.com/datalad/datalad-registry/blob/68e3d459edcdda51d5836527e9dfef325c98cde6/.github/workflows/type_check.yml#L3) and different [environment](https://github.com/datalad/datalad-registry/blob/68e3d459edcdda51d5836527e9dfef325c98cde6/.github/workflows/type_check.yml#L18-L19). Because of the removal of the `mypy run`, I also removed the no-longer-needed `pip install mypy types-requests types-PyYAML` to speed up the `Tests` workflow.

PR #392, opened 2025-09-05
Add column for last commit date, allow for subset of columns to be shown, overhaul sorting

yarikoptic:
made with claude

not tested locally yet


yarikoptic:
it had been too long, I will need to redo it

Issue #407, opened 2026-06-16
Establish replication to neurodesk people server


yarikoptic:
@iishiishii ATM is torturing our API to fetch the entire DB into Australia somewhere and it goes slow.  We should produce an easy to import dump and for such people as @iishiishii just to add to our replication from typhon.

candleindark:
I think what's being asked for here is a periodically published dump of the public `/api/v2/dataset-urls` output (all pages aggregated into one file), so a consumer can grab a single file instead of paging the API hundreds of times. That matches how @iishiishii is already consuming the data, carries no coupling to our internal DB schema, and is safe to publish publicly. I'd recommend we go with that.

For completeness, two alternatives I considered but would not pursue:

<details>
<summary>Alternatives and why not</summary>

- **A full `pg_dump` of the database.** This would let a consumer restore a byte-identical DB, but it exposes internal/operational columns (cache paths, check-scheduling and failure-count bookkeeping, etc.) that are deliberately not part of the public API, and it ties consumers to our exact schema and migration state.
- **Having a third party stand up a streaming replica of our primary.** Technically possible via our read-only mode, but it's hard to coordinate: it requires primary-side network/role/config changes per replica and keeping Postgres versions in lockstep, so it's not really self-service.

</details>

@yarikoptic does the aggregated `/api/v2/dataset-urls` dump match what you had in mind? If so I'll look into producing and publishing it on a schedule.

yarikoptic:
I thought of what you called **third party stand up a streaming replica of our primary** but was not aware of technical detail/complications.  I feel it might still be feasible via containerization and succinct config to adjust whatever needed to be adjusted on primary-side.   But may be indeed we should assess first of how much torturing is done ATM. @iishiishii how often you do your "replication" and how long it takes ATM?

iishiishii:
Thanks for starting the convo! I first thought the issue comes from using ILIKE for searching in very large pool so I experimented trigram index, it was much faster for metadata and url but not everything. The aggregated /api/v2/dataset-urls sounds like a good idea for the current bottleneck when it fetches all pages and search so it doesn't keep hitting the API. I will come back with better benchmarking in a week after sorting out my thesis and viva :)

candleindark:
I'd suggest we settle on two things and drop the third-party replica idea:

1. Add an endpoint that returns all dataset URLs in one response (the public serialized fields only, so no internal columns leak), cached or not as needed.
[...]

PR #412, opened 2026-07-24
feat: add endpoint to download all dataset URLs as a JSON Lines file

candleindark:
Toward #407. Adds a single-file bulk download of all dataset URLs so a consumer can grab one file instead of paging `GET /api/v2/dataset-urls` hundreds of times.

## New endpoint: `GET /api/v2/dataset-urls/all`

Returns every dataset URL as a **JSON Lines** (`.jsonl`) file, one JSON object per line. Each line has the same shape as an element of the `dataset_urls` array of the paginated endpoint (a `DatasetURLRespModel`), so no internal database columns are exposed. A binary `return_metadata` flag (default `false`), the counterpart of the tri-state `return_metadata` query parameter of the paginated endpoint, controls whether each line includes the `metadata` field populated by content.

<details>
<summary>Response details and behavior</summary>

- `Content-Type: application/x-ndjson`, served as a download via `Content-Disposition: attachment; filename=dataset-urls.jsonl`.
- Without `return_metadata` (default), each line omits the `metadata` field.
- With `return_metadata=true`, each line includes `metadata` as a list of metadata objects by content, the same content form as `return_metadata=content` on the paginated endpoint.
- The response is streamed with `yield_per` so the full set of records is not serialized into memory at once, and metadata is eagerly loaded via `selectinload` when requested to avoid an N+1 query.
- The path `/all` does not collide with the existing `/<int:id>` route, since static rules take precedence over converter rules in Werkzeug.

</details>

<details>
<summary>Scope: what this intentionally leaves out</summary>

- **No filter/search parameters.** Unlike the paginated endpoint, the dump takes only `return_metadata`; it is always a full snapshot of all dataset URLs.
- **No caching or scheduled publishing.** The thread floats caching the pre-serialized gzipped bytes and publishing a dump on a schedule. Deferred as a potential follow-up: the plan is to test this endpoint in operation first, and pursue caching/publishing only if the current setup's performance is not acceptable.

</details>

## Test plan

- [x] `black`, `isort`, `flake8`, `mypy` clean on the changed files.
- [x] Full `test_dataset_urls.py` module passes (148 tests, 9 new in `TestAllDatasetURLs`).
- [x] OpenAPI spec documents the endpoint (the `application/x-ndjson` 200 response and the `return_metadata` parameter both appear at `/openapi/openapi.json`).
- [ ] Exercise the endpoint against a production-scale dataset to confirm performance is acceptable before considering the caching/publishing follow-up.

<details>
<summary>New test coverage (<code>TestAllDatasetURLs</code>)</summary>

- Empty database returns an empty body with the `application/x-ndjson` content type.
- `Content-Disposition` offers the response as `dataset-urls.jsonl`.
- Without `return_metadata`: all dataset URLs are returned and no line has a `metadata` field.
- With `return_metadata=true`: each line includes `metadata` by content, with the expected per-URL counts.
- A non-boolean `return_metadata` value is rejected with `422`.

</details>

### DIFF
--- a/datalad_registry/blueprints/api/dataset_urls/__init__.py
+++ b/datalad_registry/blueprints/api/dataset_urls/__init__.py
@@ -1,16 +1,18 @@
 # This file is for defining the API endpoints related to dataset URls
 
+from collections.abc import Iterator
 from json import loads
 import operator
 from pathlib import Path
 
 from celery import group
-from flask import abort, current_app, url_for
+from flask import abort, current_app, stream_with_context, url_for
 from flask_openapi3 import APIBlueprint, Tag
 from lark.exceptions import GrammarError, UnexpectedInput
 from psycopg2.errors import UniqueViolation
 from sqlalchemy import ColumnElement, and_, select
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm import selectinload
 
 from datalad_registry.models import RepoUrl, db
 from datalad_registry.search import parse_query
@@ -27,6 +29,7 @@ from .models import (
     DatasetURLRespBaseModel,
     DatasetURLRespModel,
     DatasetURLSubmitModel,
+    DumpQueryParams,
     MetadataReturnOption,
     OrderKey,
     PathParams,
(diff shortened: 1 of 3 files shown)


datalad_registry/blueprints/api/dataset_urls/__init__.py
```diff
@@ -362,6 +365,64 @@ def cache_path_trans(cache_path: Path) -> str:
...
+    responses={
+        "200": {
+            "description": "A JSON Lines (`.jsonl`) file in which each line is a JSON "
+            "object representing a dataset URL, with the same fields as an element of "
+            "the `dataset_urls` array returned by the `GET /api/v2/dataset-urls` "
+            "endpoint. Each line optionally includes the `metadata` of the dataset at "
+            "the URL, depending on the `return_metadata` flag.",
+            "content": {
+                "application/x-ndjson": {
+                    "schema": {"type": "string", "format": "binary"}
```
yarikoptic, line 379: binary?

datalad_registry/blueprints/api/dataset_urls/__init__.py
```diff
@@ -362,6 +365,64 @@ def cache_path_trans(cache_path: Path) -> str:
...
+                    **DatasetURLRespBaseModel.from_orm(repo_url).dict(),
+                    metadata_=None,
+                )
+            yield line_model.json(exclude_none=True) + "\n"
+
+    resp = current_app.response_class(
+        stream_with_context(gen_lines()),
+        mimetype="application/x-ndjson",
+    )
+    resp.headers["Content-Disposition"] = "attachment; filename=dataset-urls.jsonl"
```
yarikoptic, line 422: might be worth right away to compress it but better be done along with caching so we do not overburden the server. And then see what mime type etc to provide back for that `.jsonl.gz`.

For when caching implemented, make sure for atomic rename of the (re)generated file into the place where served from.

yarikoptic (changes requested):
Let's implement celery task and thus caching right away, because if this is implemented and deployed, could put too much stress on the server.  When caching implemented, just make API call return a redirect to target file, and ideally collect and fill out relevant to cached data information such as ETag (md5sum IIRC would suffice) and potentially other fields (like expiration date and when was generated etc).

datalad_registry/blueprints/api/dataset_urls/__init__.py
candleindark: `format: binary` is correct as long as we are on OpenAPI 3.0. `flask-openapi3` hardcodes `openapi_version = "3.0.3"`, and in
 3.0 `{"type": "string", "format": "binary"}` is the idiomatic way to declare an opaque file payload, which is what makes Swagger UI offer a download rather than trying to rende
r a schema. It would need to become `contentMediaType` if we ever move to 3.1. I will add a comment saying so.

Nevertheless, this will most likely be replaced by the new design/implementation.

candleindark:
> Let's implement celery task and thus caching right away, because if this is implemented and deployed, could put too much stress on the server. When caching implemented, just make API call return a redirect to target file, and ideally collect and fill out relevant to cached data information such as ETag (md5sum IIRC would suffice) and potentially other fields (like expiration date and when was generated etc).

Agreed on doing the Celery task and caching up front, and on having the API return a redirect. Implementing that raises an infrastructure question since it depends on what the servers can host.

**The problem:** the worker that would generate the dump runs on the primary, but the public-facing instance is the read-only replica. The two share no filesystem, so there is currently nowhere for the worker to write a file that the public instance can serve.

**What I would like:** a location the primary can upload/write the generated dump to, and that is served publicly.

candleindark:
Note:

The follow query shows that, before gzip, jsonl file is about 9.6 MB without metadata and 70 MB with metadata.

```
dlreg=# WITH slim AS (
  SELECT r.id, jsonb_build_object(
           'url', r.url, 'id', r.id, 'ds_id', r.ds_id,
           'head_describe', r.head_describe,
           'annex_key_count', r.annex_key_count,
           'annexed_files_in_wt_count', r.annexed_files_in_wt_count,
           'annexed_files_in_wt_size', r.annexed_files_in_wt_size,
           'last_update_dt', r.last_update_dt,
           'git_objects_kb', r.git_objects_kb,
           'processed', r.processed, 'last_chk_dt', r.last_chk_dt) AS obj
  FROM repo_url r WHERE r.processed
), meta AS (
  SELECT m.url_id, jsonb_agg(jsonb_build_object(
           'extractor_name', m.extractor_name,
           'dataset_describe', m.dataset_describe,
           'dataset_version', m.dataset_version,
           'extractor_version', m.extractor_version,
           'extraction_parameter', m.extraction_parameter,
           'extracted_metadata', m.extracted_metadata)) AS arr
  FROM url_metadata m GROUP BY m.url_id
)
SELECT count(*) AS n_lines,
       pg_size_pretty(sum(octet_length(s.obj::text) + 1)) AS slim_jsonl,
       pg_size_pretty(sum(octet_length(
         (s.obj || jsonb_build_object('metadata', coalesce(mt.arr, '[]'::jsonb)))::text) + 1
       )) AS full_jsonl
FROM slim s LEFT JOIN meta mt ON mt.url_id = s.id;
 n_lines | slim_jsonl | full_jsonl
---------+------------+------------
   24792 | 9809 kB    | 70 MB
(1 row)
```

PR #415, opened 2026-08-13
fix: switch away from `git-annex` (or stale `master`) after clone

### LINKED ISSUE #414: Apparently we might need manual step to figure out proper branch
This is to workaround
- https://github.com/OpenNeuroOrg/openneuro/issues/4037

and overall I guess creation of github siblings and pushing both branches at the same time which then might result in git-annex being the default.

Here in datalad-registry we seems to need to come up with a strategy to

- clone the URL
- see if that is git-annex branch and if so -- switch to other 'main' or 'master'
- if both 'main' and 'master' branches present (one might be checked out) -- check which one is more recent and check that one out

this way we can do generally without relying on querying github like we ending up doing in OpenNeuroStudies

yarikoptic:
Some origins — notably several `OpenNeuroDatasets` and `nemarDatasets` repos — advertise `git-annex` as their default branch, or keep `master` as default while `main` holds the newer content. `datalad clone` follows the advertised HEAD, so we end up extracting metadata from the wrong branch (e.g. NeMar datasets showing up in the registry without BIDS metadata because `git-annex` was checked out).

Add `pick_preferred_branch` / `ensure_preferred_branch_checked_out` in `datalad_tls`: after clone, prefer whichever of `main`/`master` has the more recent commit at origin (or the sole one that exists), check it out, and repoint local `origin/HEAD` so downstream code sees the fix. Wire the fixup in at both clone sites — `process_dataset_url` and `update_ds_clone::reclone_ds` — and swap the update-path branch comparison to use the preferred branch so we reclone when the primary branch changes (e.g. `main` overtakes `master`).

- Closes #414


datalad_registry/utils/datalad_tls.py
```diff
@@ -152,6 +155,49 @@ def get_origin_default_branch(ds: Dataset) -> str:
...
+        name: info
+        for name, info in get_origin_branches(ds).items()
+        if name in _PREFERRED_BRANCH_CANDIDATES
+    }
+    if not candidates:
+        return None
+    return max(
+        candidates,
+        key=lambda name: datetime.fromisoformat(candidates[name]["last_commit_dt"]),
+    )
```
candleindark, line 173: `datetime.fromisoformat` cannot parse a `Z` suffix before Python 3.11, and `%(authordate:iso8601-strict)` renders UTC commits with exactly that suffix. Any repo whose `main`/`master` tip was authored in UTC therefore raises `ValueError` here. This is what is currently making the `test` job red (`26 failed, 430 passed, 5 errors`).

There is no supported environment where this works: `setup.cfg` declares `python_requires = >= 3.9`, all three workflows pin `3.9`, and the runtime image `phusion/baseimage:jammy-1.0.1` carries 3.10.

The consequence is not a mis-ranking but a hard failure in both task paths. In `process_dataset_url` the exception propagates and the fresh clone is deleted, so the URL is never processed. In `chk_url_to_update` it lands in the handler that does `n_failed_chks += 1`, and once that reaches `DATALAD_REGISTRY_MAX_FAILED_CHKS_PER_URL` the dispatcher's `RepoUrl.n_failed_chks < max_failed_chks` filter stops selecting the URL for good.

Since nothing outside `datalad_tls.py` reads `last_commit_dt` (the API blueprints and templates only test JSONB *keys*, as in `branches ? 'git-annex'`), the tidiest fix is in `get_origin_branches` rather than here, so every stored value is uniformly parseable: either normalize the suffix, or record `%(authordate:unix)` and compare integers. Worth noting that rows written before the fix keep the `Z` spelling until their next update, so the column holds both spellings for a while unless they are rewritten deliberately.

<details>
<summary>Verification</summary>

Git's rendering, confirmed with git 2.55.0. Only UTC is affected:

```console
$ git for-each-ref --format='%(refname) [%(authordate:iso8601-strict)]' refs/heads/
refs/heads/main [2020-01-01T00:00:00Z]        # commit authored +00:00
refs/heads/main [2020-01-01T00:00:00+02:00]   # commit authored +02:00
```

On the project interpreter (3.9.23):

```console
$ python -c "from datetime import datetime; datetime.fromisoformat('2020-01-01T00:00:00Z')"
ValueError: Invalid isoformat string: '2020-01-01T00:00:00Z'
```

And from the PR's own CI run:

```
E   ValueError: Invalid isoformat string: '2026-08-13T14:17:21Z'
datalad_registry/utils/datalad_tls.py:172: ValueError
======= 26 failed, 430 passed, 5 warnings, 5 errors in 112.44s (0:01:52) =======
```

</details>

datalad_registry/utils/datalad_tls.py
```diff
@@ -152,6 +155,49 @@ def get_origin_default_branch(ds: Dataset) -> str:
...
+        # working tree, which blocks a plain checkout.
+        ds.repo.call_git(["checkout", "-f", "-B", preferred, f"origin/{preferred}"])
+
+    ds.repo.call_git(
+        [
+            "symbolic-ref",
+            "refs/remotes/origin/HEAD",
+            f"refs/remotes/origin/{preferred}",
+        ]
+    )
```
candleindark, line 198: Worth adding a `ds.config.reload()` at the end of this function, because without it the PR stops one line short of fixing `ds_id` for the very repos it targets.

`Dataset.id` in datalad 0.19.6 is just `self.config.get('datalad.dataset.id', None)`, with no reload of its own. `Dataset.config` delegates to `repo.config`, a `ConfigManager` held by the flyweight repo object and populated during `dl.clone`, at which point the working tree still held the `git-annex` branch and so had no `.datalad/config`. Checking out `main` here materializes that file on disk, but nothing re-reads it, so the manager keeps answering from its clone-time snapshot and `_update_dataset_url_info` records `ds_id = None` again.

To be clear, this is not a regression: these repos already stored `ds_id = NULL` before the PR, for the same reason. It is simply an opportunity that is now within reach and currently missed. As it stands the PR corrects the branch and the extracted metadata but still records the dataset as not being a DataLad dataset.

There is a second benefit. `get_pure_annex_ds_collection_stats` classifies a dataset as *pure annex* (an annex repo that is not a DataLad dataset) with `branches ? 'git-annex' AND ds_id IS NULL`. Every repo this PR targets has a `git-annex` branch by construction, so they are all currently miscounted as pure annex in the overview page's collection stats. Reloading the config would correct that statistic on the next update too.

<details>
<summary>Reproduction (datalad 0.19.6)</summary>

An annex source dataset with `HEAD` pointed at `git-annex`, mimicking the issue #414 failure mode, then cloned:

```
branch after clone:                 git-annex
ds.id right after clone:            None      # <- also the pre-PR state
.datalad/config on disk:            True      # after checkout -f -B main origin/main
ds.id after checkout (no reload):   None
ds.id after config.reload():        9ec93277-3124-497c-8eae-d01a69a4fee8
```

</details>

datalad_registry/tasks/utils/__init__.py
```diff
@@ -128,14 +134,17 @@ def reclone_ds() -> Dataset:
...
 
-    # The current default branch of the origin remote, the copy of the dataset
-    # located at the given URL, that the local clone is tracking
-    current_origin_default_branch = get_origin_default_branch(current_ds_clone)
+    # Prefer main/master over origin's advertised default (may be `git-annex`
+    # or an outdated `master`). See
+    # https://github.com/datalad/datalad-registry/issues/414
+    target_branch = pick_preferred_branch(
+        current_ds_clone
+    ) or get_origin_default_branch(current_ds_clone)
```
candleindark, line 142: The `or get_origin_default_branch(...)` fallback only fires when **neither** `main` nor `master` exists at origin. Whenever either ref is present, `pick_preferred_branch` returns a name, the `or` short-circuits, and origin's advertised default is never consulted at all (`get_origin_default_branch` is the only caller of `git ls-remote --symref origin HEAD`).

The choice then rests entirely on `refs/remotes/origin/*`, the local remote-tracking refs, which the `git fetch` above does not prune.

datalad_registry/utils/datalad_tls.py
candleindark, line 173: This overrides origin's default even when that default is a legitimate third branch, which is wider than issue #414 asks for.

Take a repo whose real default is `dev` or `trunk` but which still carries a leftover `main` or `master`. The leftover branch is the one that gets checked out, in `process_dataset_url` at registration and in `reclone_ds` on the update path. The registry then tracks a branch the repo owner did not choose, where today it tracks the right one. Worse, that branch never moves, so the entry silently freezes: the fast-forward path finds nothing and the record stays frozen while the real default advances.

datalad_registry/utils/datalad_tls.py
```diff
@@ -152,6 +155,49 @@ def get_origin_default_branch(ds: Dataset) -> str:
...
+
+def ensure_preferred_branch_checked_out(ds: Dataset) -> None:
+    """
+    Check out `pick_preferred_branch(ds)` and repoint local `origin/HEAD` at
+    it, so downstream reads of `origin/HEAD` see the corrected default.
+    No-op if no preferred branch exists at origin.
+    """
+    preferred = pick_preferred_branch(ds)
+    if preferred is None:
+        return
```
candleindark, line 184: @yarikoptic, a question on how to read the strategy in issue #414. It lists:

> - see if that is git-annex branch and if so -- switch to other 'main' or 'master'
> - if both 'main' and 'master' branches present (one might be checked out) -- check which one is more recent and check that one out

Is the second bullet meant as a subcondition of the first, so that it applies only once the default has been found to be `git-annex`? Or is it independent, so that a repo whose advertised default is a real `main` or `master` should be re-examined too?

The reason it matters: as implemented the switch is unconditional. Whenever origin has a `main` or a `master`, the clone is moved onto it, whatever origin advertises as its default, and `refs/remotes/origin/HEAD` is repointed to match. A repo whose default is `dev` and which still carries a long-abandoned `master` ends up checked out on that `master`.

If the second bullet is a subcondition, the logic should trigger only when origin's advertised default is `git-annex`, and every other repo should be left on whatever origin advertises.

<details>
<summary>Demonstration: origin advertises <code>dev</code>, clone ends up on a stale <code>master</code></summary>

A source repo whose default branch is `dev` (2025), carrying a leftover `master` last touched in 2019, cloned and then passed through `ensure_preferred_branch_checked_out`:

```
origin advertises default : dev
branch after clone        : dev
pick_preferred_branch()   : master
branch after ensure(...)  : master
origin/HEAD now points at : refs/remotes/origin/master
files in working tree     : ['.noannex', 'seed.txt']
```

`newer.txt`, added on `dev`, is gone from the working tree, so metadata extraction would run against the 2019 content. Non-UTC author dates were used here so that the `Z`-suffix parsing problem stays out of the way.

</details>

datalad_registry/utils/datalad_tls.py
```diff
@@ -152,6 +155,49 @@ def get_origin_default_branch(ds: Dataset) -> str:
...
+    sole one that exists, or `None`.
+    """
+    candidates = {
+        name: info
+        for name, info in get_origin_branches(ds).items()
+        if name in _PREFERRED_BRANCH_CANDIDATES
+    }
+    if not candidates:
+        return None
+    return max(
```
candleindark, line 170: When `main` and `master` point at the same commit, the two dates are equal and `max` returns the first maximal element, that is, whichever key `get_origin_branches` yielded first. That happens to be `main`, since `git for-each-ref` sorts by refname and dicts preserve insertion order. The right branch wins, but by accident rather than by intent.

The tie is not an exotic case: it is the normal state right after a repo renames `master` to `main` and keeps both names pointing at the same commit.

An explicit tie-break, plus a comment saying which branch is meant to win, would keep this from flipping if `get_origin_branches` ever changes its ordering.

<details>
<summary>Current behavior</summary>

```
origin branches, in iteration order:
   main     25ff4989  2020-01-01T00:00:00+02:00
   master   25ff4989  2020-01-01T00:00:00+02:00
tie is real       : True
pick_preferred_branch(): main
```

</details>

datalad_registry/tasks/utils/__init__.py
```diff
@@ -128,14 +134,17 @@ def reclone_ds() -> Dataset:
 
     current_ds_clone.repo.call_git(["fetch"])
```
candleindark, line 135: Without `--prune`, this leaves remote-tracking refs behind for branches that no longer exist at origin, and `pick_preferred_branch` reads exactly those refs.

Adding `--prune` here would also stop the `branches` column from recording branches that have been deleted at origin.

datalad_registry/utils/datalad_tls.py
```diff
@@ -152,6 +155,49 @@ def get_origin_default_branch(ds: Dataset) -> str:
...
+        candidates,
+        key=lambda name: datetime.fromisoformat(candidates[name]["last_commit_dt"]),
```
candleindark, line 172: `last_commit_dt` carries `authordate`, which is preserved by rebase, `git am`, cherry-pick, and history imports, so it does not record when a branch last moved. A `main` whose tip was applied from an old patch can lose to a `master` abandoned years ago, and the abandoned branch is then the one checked out.

`committerdate` is the field that answers "which branch moved most recently", and the docstring above already promises that behavior ("more-recently-committed").

The field comes from `get_origin_branches`, which also feeds the `branches` column, so exposing `committerdate` there alongside `authordate` is probably cleaner than swapping one for the other.

datalad_registry/utils/datalad_tls.py
```diff
@@ -152,6 +155,49 @@ def get_origin_default_branch(ds: Dataset) -> str:
...
+    """
+    Check out `pick_preferred_branch(ds)` and repoint local `origin/HEAD` at
+    it, so downstream reads of `origin/HEAD` see the corrected default.
+    No-op if no preferred branch exists at origin.
+    """
+    preferred = pick_preferred_branch(ds)
+    if preferred is None:
+        return
+
+    current_branch = ds.repo.call_git(["symbolic-ref", "--short", "HEAD"]).strip()
```
candleindark, line 186: `symbolic-ref --short HEAD` exits non-zero on a detached HEAD, which datalad turns into a `CommandError`.

A top-level `datalad clone` always leaves HEAD attached, so this is not reachable today at either call site. It is only worth raising because of what an exception costs in this particular spot: in `chk_url_to_update` it increments `n_failed_chks`, and once that reaches `DATALAD_REGISTRY_MAX_FAILED_CHKS_PER_URL` the dispatcher stops selecting the URL for good.

`git branch --show-current` prints an empty string rather than failing, if you want the guard for free.

candleindark (changes requested):

